PR #7, opened 2018-10-31
Add prometheus monitoring client on readers

ashangit:


readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/metrics/PrometheusHttpConsumerMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+            .name("garmadon_reader_metrics").help("Garmadon reader metrics")
+            .labelNames("name", "hostname", "release")
+            .register();
+
+    private static final Gauge baseKafkaMetricsGauge = Gauge.build()
+            .name("garmadon_kafka_metrics").help("Kafka producer metrics")
+            .labelNames("name", "hostname", "release")
+            .register();
+
+    public HTTPServer server;
```
Willymontaz: can you put this instance field just before the constructor ?

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/metrics/PrometheusHttpConsumerMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static ScheduledThreadPoolExecutor scheduler = new ScheduledThreadPoolExecutor(1);
+
+    private static ObjectName oname;
```
Willymontaz: use camel case

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/metrics/PrometheusHttpConsumerMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+    static {
+        // Expose JMX, GCs, classloading, thread count, memory pool
+        DefaultExports.initialize();
+        try {
+            oname = new ObjectName("kafka.consumer:type=consumer-metrics,client-id=" + GarmadonReader.CONSUMER_PREFIX_NAME + "." + GarmadonReader.HOSTNAME);
```
Willymontaz: GarmadonReader.CONSUMER_PREFIX_NAME + "." + GarmadonReader.HOSTNAME -> code already used somewhere else. I propose to dedicate a property for it

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
```diff
@@ -11,20 +13,37 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(GarmadonReader.class);
+    public static final String CONSUMER_PREFIX_NAME = "garmadon.reader";
+
+    public static String HOSTNAME;
+
+    static {
+        try {
+            HOSTNAME = InetAddress.getLocalHost().getCanonicalHostName();
+        } catch (UnknownHostException e) {
+            LOGGER.error("", e);
```
Willymontaz: Then hostname is null which is dangerous for a public field. Either we set a default value or we must exit(1) if we cannot get it.

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/metrics/ConsumerMetrics.java
```diff
@@ -0,0 +1,5 @@
+package com.criteo.hadoop.garmadon.reader.metrics;
+
+public interface ConsumerMetrics {
+    void stop();
```
Willymontaz: I find that this interface is not clearly obvious and is only the result of using 2 implementations. In other words, just having a close method seems to not say much about ConsumerMetrics. I would have expected such an interface to also have a start method because starting everything via statics is only a preference of PrometheusHttpConsumerMetrics
Maybe too early to extract an interface. Another way to approach this in terms of design can be to have a DefaultConsumerMetrics, that do nothing and the PrometheusHttpConsumerMetrics would extend this default class

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/metrics/PrometheusHttpConsumerMetrics.java
Willymontaz: When this is evaluated, are we sure the consumer already setup the jmx ?

ashangit: Here we are just creating the ObjectName object not looking at the MBean
Then we look at the MBean on exposeKafkaMetrics scheduled every 30secs so there is no issue if first time "kafka.consumer:type=consumer-metrics,client-id=" + GarmadonReader.CONSUMER_PREFIX_NAME + "." + GarmadonReader.HOSTNAME"
 is not readable in MBean

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/metrics/ConsumerMetrics.java
ashangit: Don't see the issue or whet would be improved with DefaultConsumerMetrics instead of NoneConsumerMetrics (I can rename it if you want)
I will add start method to make it more "generic"

Willymontaz: The problem is not the name. The way the interface looked just shows that the reader class has very little control over ConsumerMetrics (just stopping it and this is only usefull for one of the implementations), and indeed, I think that there is no need to add the ConsumerMetrics to the reader class. If we want to properly stop Prometheus metrics we can register to the CompletableFuture returned by the stop() method of the reader and close the Prometheus stuff with whenComplete.

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/metrics/PrometheusHttpConsumerMetrics.java
Willymontaz: ok

PR #9, opened 2018-11-02
[BREAKING] - Provide timestamp in Garmadon's ProtocolMessage

ashangit:
Also remove duplicate timestamp field from protobuf events

JIRA:LAKECOMP-878

### DIFF
--- a/frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
+++ b/frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
@@ -1,5 +1,6 @@
 package com.criteo.hadoop.garmadon.spark.listener;
 
+import com.criteo.hadoop.garmadon.TriConsumer;
 import com.criteo.hadoop.garmadon.event.proto.SparkEventProtos;
 import com.criteo.hadoop.garmadon.schema.enums.State;
 import com.criteo.hadoop.garmadon.schema.events.Header;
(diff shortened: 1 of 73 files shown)


schema/src/main/java/com/criteo/hadoop/garmadon/protocol/ProtocolVersion.java
```diff
@@ -4,7 +4,7 @@
 import java.util.List;
 
 public class ProtocolVersion {
-    public static final byte VERSION = 1;
+    public static final byte VERSION = 3;
```
Willymontaz: Seems like this is a hack to prevent authorizing this release. Shouldn't we instead manager the AUTHORIZED release by hand and explicitly list still managed versions ?

frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
```diff
@@ -190,14 +188,17 @@ public void onTaskEnd(SparkListenerTaskEnd taskEnd) {
         try {
             String status = taskEnd.taskInfo().status();
 
+            Header taskHeader = header.cloneAndOverride(Header.newBuilder()
```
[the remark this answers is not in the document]
gray-matter: Looks like a piece of code we discussed together and sneaked into this review :-)

schema/src/main/java/com/criteo/hadoop/garmadon/protocol/ProtocolVersion.java
ashangit: I don't think that managing release manually is a good  choice

Willymontaz: How is it worse than a piece of code forcing to accept the previous version and hacking it the first time we use it ?

Willymontaz: To add clarity on "managing by hand", I was meaning to populate the list line 10 with actual version supported and not necessarily the previous one.

PR #15, opened 2018-11-09
Manage ingestion issue with proto2 event being badly ingest

ashangit:
Also correct little issue about release naming in prometheus and it's usage in graphite

### DIFF
--- a/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
+++ b/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
@@ -12,6 +12,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.nio.BufferUnderflowException;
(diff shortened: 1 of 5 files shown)


readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
```diff
@@ -129,18 +146,16 @@ protected void readMsg() {
...
                                 } catch (DeserializationException e) {
-                                    LOGGER.debug("Cannot deserialize event from kafka record " + record + " with type " + typeMarker);
+                                    LOGGER.debug("Cannot deserialize event from kafka record {} with type {}", record, typeMarker);
+                                    break;
                                 }
                             }
-
                             CommittableOffset<String, byte[]> committableOffset = new CommittableOffset<>(consumer, record.topic(), record.partition(), record.offset());
 
-                            if (header != null && body != null) {
```
Willymontaz: We can leave this line for safety but still use the break you introduced to exit the filter loop

PR #16, opened 2018-11-12
Create script to easily bump release

ashangit:


scripts/bump_release.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+set -e
+set -x
+
+NEW_GARMADON_RELEASE=$1
```
gray-matter: Fail if $# < 0 maybe?

scripts/bump_release.sh
```diff
@@ -0,0 +1,20 @@
...
+cd "`dirname "$0"`/.."
+
+# Create/Switch to release branch locally
+git checkout release_$NEW_GARMADON_RELEASE || git checkout -b release_$NEW_GARMADON_RELEASE
```
gray-matter: What happens if both fail? Shouldn't we exit the script?

scripts/bump_release.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Bump release
+mvn versions:set -DnewVersion=${NEW_GARMADON_RELEASE}
+find . -name pom.xml.versionsBackup -delete
+sed -i "s#<garmadon.version>.*</garmadon.version>#<garmadon.version>${NEW_GARMADON_RELEASE}</garmadon.version>#g" pom.xml
+
+# Commit release
+git add *pom.xml
```
gray-matter: Unnecessary star?

scripts/bump_release.sh
```diff
@@ -0,0 +1,20 @@
...
+git commit -a -m "Prepare release ${NEW_GARMADON_RELEASE}"
```
gray-matter: What's the point of git add if you then git commit -a?

scripts/bump_release.sh
ashangit: set -e -> Exit immediately if a command exits with a non-zero status.

gray-matter: True that!

PR #20, opened 2018-11-15
Discard messages from Agent when protocol version is incorrect instead of closing the connection

Willymontaz:


ashangit:
Please add a TODO indicating that once the migration to new agent release is performed (meaning once all JVMs have been restarted on new agent release) we should revert that commit

PR #26, opened 2018-11-19
Refacto on jvm-statistics modules

ashangit:


jvm-statistics/protobuf/src/test/java/com/criteo/jvm/ProtobufGCNotificationsTest.java
```diff
@@ -12,25 +15,36 @@
 
 public class ProtobufGCNotificationsTest {
 
-    private static final Pattern GC_PATTERN = Pattern.compile("collector_name: \".*\"\\s+" +
-            "pause_time: \\d+\\s+" +
-            "cause: \"System.gc\\(\\)\"\\s+" +
-            ".*" +
```
jpbempel: I have removed the matching on eden & survivor because sometimes does spaces are not changing when triggering the FullGC and may not appear in the logs

ashangit: In fact some times they are no change in it but with JsonFormat.printer().includingDefaultValueFields() the output json always contains all values define on the proto schema even if equal to default (0)

PR #27, opened 2018-11-19
Compute severity based on number of created files

mlachkar:


readers/heuristics/src/main/java/com/criteo/hadoop/garmadon/heuristics/FileHeuristic.java
```diff
@@ -55,7 +56,15 @@ public void onContainerCompleted(String applicationId, String attemptId, String
     @Override
     public void onAppCompleted(String applicationId, String attemptId) {
         //TODO compute severity based on number of deleted, renamed, read, written...
-        HeuristicResult result = new HeuristicResult(applicationId, attemptId, FileHeuristic.class, HeuristicsResultDB.Severity.NONE, HeuristicsResultDB.Severity.NONE);
+        int created = written.forApp(applicationId, attemptId).getCount() -
```
Willymontaz: does that mean that creating 1 000 000 000 files and deleting 1 000 000 000 is fine ? It looks like you should only consider the written number of files.

Willymontaz:
Please make it tunable via properties rather than hardcoded thresholds

PR #39, opened 2018-11-21
Add Checkstyle and start implementing some syntax checks

ashangit:


checkstyle.xml
```diff
@@ -0,0 +1,134 @@
...
+<module name="Checker">
+    <module name="JavadocPackage"/>
+    <module name="TreeWalker">
+        <!-- Checks for Annotation. -->
+        <!--
+        <module name="AnnotationLocation"/>
+        <module name="AnnotationOnSameLine"/>
+        <module name="AnnotationUseStyle"/>
+        <module name="MissingDeprecated"/>
+        <module name="MissingOverride"/>
```
Willymontaz: We can enable MissingOverrides

checkstyle.xml
```diff
@@ -0,0 +1,134 @@
...
+    - some best practices
+  Checkstyle is very configurable. Be sure to read the documentation at
+  http://checkstyle.sf.net (or in your downloaded distribution).
+  Most Checks are configurable, be sure to consult the documentation.
+  To completely disable a check, just comment it out or delete it from the file.
+  Finally, it is worth reading the documentation.
+-->
+
+<module name="Checker">
+    <module name="JavadocPackage"/>
```
Willymontaz: Can we remove the rule JavadocPackage (forcing package-info.java) ? If we don't do javadocs it is useless

ashangit: We don't do so far but I think that we must do
I would prefer to keep it as a startup and then add more rules about javadoc
Is it fine or do you prefer to not have javadoc at all?

checkstyle.xml
ashangit: Checks on annotation has not being impelemented so far look at commit messages to see which "checks" have been implemented
I will add in an other commit (with some other checks still not implemented)

checkstyle.xml
Willymontaz: Ok for looking forward to add more doc in the code. package-info.java is not that used but I don't really care to remove it or leave it.

PR #41, opened 2018-11-22
Get canonical hostname to ensure having same hostname format

ashangit:
Hostname from containers based on NM env variable is in canonical format
the one from standalone JVM is not leading to different format


ashangit:
This is in fact not breaking and it can wait for release 1.1.0 to be released
There is nothing breaking here just a format alignment of hostname

PR #43, opened 2018-11-22
Correct bump_release.sh for OSX

Willymontaz:
Closes https://github.com/criteo/garmadon/issues/42

### DIFF
--- a/scripts/bump_release.sh
+++ b/scripts/bump_release.sh
@@ -26,7 +26,11 @@ git checkout release_$NEW_GARMADON_RELEASE || git checkout -b release_$NEW_GARMA
 # Bump release
 mvn versions:set -DnewVersion=${NEW_GARMADON_RELEASE}
 find . -name pom.xml.versionsBackup -delete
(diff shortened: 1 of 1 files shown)


scripts/bump_release.sh
```diff
@@ -26,7 +26,11 @@ git checkout release_$NEW_GARMADON_RELEASE || git checkout -b release_$NEW_GARMA
 # Bump release
 mvn versions:set -DnewVersion=${NEW_GARMADON_RELEASE}
 find . -name pom.xml.versionsBackup -delete
-sed -i "s#<garmadon.version>.*</garmadon.version>#<garmadon.version>${NEW_GARMADON_RELEASE}</garmadon.version>#g" pom.xml
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
ashangit: Already told you that it won't be fine as this depends on sed version (GNU or mac one)
I'm using a mac but used the GNU release of sed, so it won't work on my computer now

PR #46, opened 2018-11-23
Change release note mechanism

Willymontaz:
Details: Fetch tags from remote, check they are only from master and correspond to a release tag


ashangit:
Doesn't work with gnu sed on mac os
```
if [[ "$OSTYPE" == "darwin"* ]]; then
   sed -i '' "s#<garmadon.version>.*</garmadon.version>#<garmadon.version>${NEW_GARMADON_RELEASE}</garmadon.version>#g" pom.xml
 else
   sed -i "s#<garmadon.version>.*</garmadon.version>#<garmadon.version>${NEW_GARMADON_RELEASE}</garmadon.version>#g" pom.xml
fi
```

PR #48, opened 2018-11-29
Second step of checkstyle implementation

ashangit:
I'm spliting in multiple step because it leads to possible lots of change and each time an other request is merged it needs to redo the works
A 3rd one will probably come after that one

### DIFF
--- a/agent/src/main/java/com/criteo/hadoop/garmadon/agent/AsyncEventProcessor.java
+++ b/agent/src/main/java/com/criteo/hadoop/garmadon/agent/AsyncEventProcessor.java
@@ -45,7 +45,7 @@ public class AsyncEventProcessor implements Runnable {
         private final Header header;
         private final Object body;
 
-        public Message(long timestamp, Header header, Object body) {
+        Message(long timestamp, Header header, Object body) {
             this.timestamp = timestamp;
             this.header = header;
             this.body = body;
--- a/agent/src/main/java/com/criteo/hadoop/garmadon/agent/headers/ContainerHeader.java
+++ b/agent/src/main/java/com/criteo/hadoop/garmadon/agent/headers/ContainerHeader.java
@@ -24,6 +24,11 @@ public final class ContainerHeader {
     private String executorId;
     private String mainClass;
 
+    private ContainerHeader() {
+        setFrameworkComponent();
+        this.header = createCachedHeader();
+    }
+
     private void setFrameworkComponent() {
         String[] commands = Utils.getArrayJavaCommandLine();
         mainClass = commands[0];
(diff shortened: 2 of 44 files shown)


jpbempel:
In Java, by the spec, any field of an object are init to ZERO, it means that boolean it's false, for ref it's is null... it's redundant. would you create a class that extends Object ? no because it's implicit and everybody knows that. Same here :)
for Map, it's very rare to have to use specific method from a HashMap, and it's a good habit to use interface instead of the concrete type because it costs nothing (perf & feature). For fields, for me this is mandatory, for local vars you could argue, but at least you don't have to think about it if you do it everywhere :)

Willymontaz:
About fields, I'm saying that sometimes you might want to be explicit. You can find a lot of classes from many respectable projects where you have int init at 0, or object to null for expressiveness.  If you talk about redundancy, we are in the same debate as doing very factorised code based on conventions vs try to express to another reader, perhaps not knowing all conventions. 
If it is not that big of a deal to set the field, and you just wanted to say "Hey, I set it here to false, that's redundant but I somehow tell you I will not do a setup method somewhere else putting it to true", I find it disappointing that a syntax tool forces you to not do it :)

"for local vars you could argue, but at least you don't have to think about it if you do it everywhere :)" -> that was exactly my point, both for maybe arguing in a philosophical layer :) but also considering this would be ok for probably the entire life of garmadon :)

Willymontaz:
About the review, 
 - how about not doing the visibility check ?
 - I let @ashangit and @gray-matter express their feeling and If we decide to keep the field initialisation rule, can we remove // Default initialization of boolean is false ?

jpbempel:
As always, style is debatable. Do you really want to enforce style then?

Willymontaz:
Yes definitely, there are rules helping readability for sure, and debating helps us to see on what we converge and what the majority of the team wants. Then we can follow the common decision. Why should it be "do" or "do not" ?

ashangit:
I agree that for fields it could be less readable even if java spec is clear, agree to remove it.
For Map I prefer to keep it as there are no need to use hashmap specific methods.
VisibilityModifier is not part of this PR unless you are speaking about an other visibility check

gray-matter:
* I would not force initializing or not initializing with default values and leave the writer decide on what's the most readable
* I don't have a strong opinion on reducing types eg. to Map. I tend to do it spontaneously, but I agree that except for method parameters, we shouldn't really care. I don't know how this is going to make the code writing/reading/maintenance any better, so I would vote against a specific rule
* As discussed w/ William, and even though I'm not sure which commit this is, I would keep the warnings around visibility, which increase understanding of the code. If reflection is an issue, then let's have dedicated unit tests to make sure that important methods have the correct visibility. I used to do this to make sure that a given class had a default constructor that seemed to be unused

ashangit:
Do we agree on removing ExplicitInitialization which enfore to not initialize boolean, don't change visibility and change IllegalType to not applied it on vars (set tokens to "ANNOTATION_FIELD_DEF,CLASS_DEF,INTERFACE_DEF,METHOD_CALL,METHOD_DEF,METHOD_REF,PARAMETER_DEF")?

ashangit:
@Willymontaz @gray-matter please review my last comment

PR #49, opened 2018-11-29
[HDFS reader] Build a shaded JAR

gray-matter:
Also remove "scope compile" from dependencies

### DIFF
--- a/readers/hdfs/pom.xml
+++ b/readers/hdfs/pom.xml
@@ -63,16 +63,6 @@
         <!-- Garmadon -->
         <dependency>
             <groupId>com.criteo.java</groupId>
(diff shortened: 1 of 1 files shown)


readers/hdfs/pom.xml
```diff
@@ -31,59 +31,108 @@
...
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-auth</artifactId>
             <version>${hadoop.version}</version>
-            <scope>compile</scope>
         </dependency>
 
         <!-- Garmadon -->
         <dependency>
```
ashangit: You can remove garmadon-jvm-statistics-protobuf and  garmadon-schema

PR #51, opened 2018-11-30
Consumer field must be public to be accessible by resourcemanager classes

ashangit:
Applying VisibilityModifier checkstyle leads to consumer being private
checkstyle is not able to detect that this field must be public as the code is injected
in resourcemanager classes at runtime


ashangit:
No this is a bug due to first commit on checkstyle which is affectiung current release

PR #56, opened 2018-12-06
Add back GC_EVENTS to ES and display GC Evengts to yarn and standalone dashboards

ashangit:
It represents around 30 % more events on ES


readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -61,7 +59,7 @@
 
     ElasticSearchReader(GarmadonReader.Builder builderReader, BulkProcessor bulkProcessorMain, String esIndexPrefix, PrometheusHttpConsumerMetrics prometheusHttpConsumerMetrics) {
         this.reader = builderReader
-                .intercept(not(hasType(GarmadonSerialization.TypeMarker.GC_EVENT)), this::writeToES)
+                .intercept(GarmadonMessageFilter.ANY.INSTANCE, this::writeToES)
```
gray-matter: Unrelated change?

ashangit: This is completely related.
Without that we still filter GC event from readers so never sending GC events to ES ;)

gray-matter: Indeed. My review skills are not that good this year :wink:

PR #57, opened 2018-12-06
[HDFS exporter] Ensure final file's parent directory exists

gray-matter:
The directory seems to be created by default with a local FS, but not with HDFS
Also factor and test the method

### DIFF
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/FileSystemUtils.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/FileSystemUtils.java
@@ -17,13 +17,11 @@ public final class FileSystemUtils {
      * @param fs            Filesystem to which these directories should belong
      * @throws IOException  When failing to create any of the directories
      */
-    public static boolean ensureDirectoriesExist(Collection<Path> dirs, FileSystem fs) throws IOException {
+    public static void ensureDirectoriesExist(Collection<Path> dirs, FileSystem fs) throws IOException {
         for (Path dir : dirs) {
             if (!fs.exists(dir) && !fs.mkdirs(dir)) {
-                return false;
+                throw new IOException(String.format("Couldn't create %s (no specific reason)", dir.toUri()));
             }
         }
-
-        return true;
     }
 }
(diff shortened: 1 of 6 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/FileSystemUtils.java
```diff
@@ -0,0 +1,29 @@
...
+     * Ensure paths exist and are directories. Otherwise create them.
+     *
+     * @param dirs          Directories that need to exist
+     * @param fs            Filesystem to which these directories should belong
+     * @throws IOException  When failing to create any of the directories
+     */
+    public static boolean ensureDirectoriesExist(Collection<Path> dirs, FileSystem fs) throws IOException {
+        for (Path dir : dirs) {
+            if (!fs.exists(dir) && !fs.mkdirs(dir)) {
+                return false;
```
jpbempel: why not throwing an IOException (or other type of exception) if you fail to create the dir?
Usage seems to abort current the process anyway. Exception message can include the right infos

jpbempel: for the log warning you can catch the IOException and log the message

gray-matter: Anything is possible, question is: is that really an issue? :-)

jpbempel: My point is to merge the usages.
Only dealing with Exception here seems the good way as this will be exceptional...

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
```diff
@@ -63,9 +64,18 @@ public Path close() throws IOException {
         writer.close();
 
         final Path finalPath = new Path(finalHdfsDir, fileNamer.computePath(dayStartTime, latestOffset));
-        fs.rename(temporaryHdfsPath, finalPath);
 
-        LOGGER.info("Committed " + finalPath.toUri());
+        if (!FileSystemUtils.ensureDirectoriesExist(Collections.singleton(finalPath.getParent()), fs)) {
+            LOGGER.warn("Couldn't ensure {} exists", finalPath.getParent());
+            return null;
```
Willymontaz: I would prefer if we can avoid returning null. I have not looked through the entire code so I don't know if we use the return value from close() method aside from tests. If not, we could return void and do testing based on behaviour (using verify) to make sure we existed the method when wanted.

gray-matter:
Updated to address both comments, please review again when you have time :-)

PR #59, opened 2018-12-07
Add process network statistics

jpbempel:
Add processrx & processtx properties that show bytes received & sent by the
sockets associated to this process or forked one.
Based on command "ss -tinp" which we parse and resolve with the list of
processes to handle forked process and sum them up to the initial JVM process

Notes:
- Handle only IPv4
- based on the refresh rate, it can miss connections that may be short lived
between 2 collections. So this is only an estimation, not accurate numbers.

### DIFF
--- a/jvm-statistics/src/main/java/com/criteo/hadoop/garmadon/jvm/statistics/NetworkStatistics.java
+++ b/jvm-statistics/src/main/java/com/criteo/hadoop/garmadon/jvm/statistics/NetworkStatistics.java
@@ -2,86 +2,216 @@ package com.criteo.hadoop.garmadon.jvm.statistics;
 
 import com.criteo.hadoop.garmadon.jvm.AbstractStatistic;
 import com.criteo.hadoop.garmadon.jvm.StatisticsSink;
-import oshi.SystemInfo;
-import oshi.hardware.NetworkIF;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.RandomAccessFile;
+import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 class NetworkStatistics extends AbstractStatistic {
     private static final String NETWORK_HEADER = "network";
-    private static final String NETWORK_RECV_SUFFIX = "_rx";
-    private static final String NETWORK_SENT_SUFFIX = "_tx";
-    private static final String NETWORK_PKT_RECV_SUFFIX = "_pktrx";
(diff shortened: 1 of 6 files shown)


jvm-statistics/src/main/java/com/criteo/hadoop/garmadon/jvm/statistics/NetworkStatistics.java
```diff
@@ -2,86 +2,216 @@
...
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 class NetworkStatistics extends AbstractStatistic {
     private static final String NETWORK_HEADER = "network";
```
ashangit: I prefer to keep it in the process section. Also we are performing some device split on readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/EventHelper.javabased on disk and network section name that we don't want to apply here

readers/elasticsearch/src/main/elasticsearch/es/template.json
```diff
@@ -756,6 +760,10 @@
           "type": "long",
           "index": false
         },
+        "processtx": {
```
ashangit: processtx -> process_network_tx

jvm-statistics/src/main/java/com/criteo/hadoop/garmadon/jvm/statistics/NetworkStatistics.java
```diff
@@ -2,86 +2,216 @@
...
+    public NetworkStatistics() {
         super(NETWORK_HEADER);
-        for (NetworkIF nic : new SystemInfo().getHardware().getNetworkIFs()) {
-            if (nic.getBytesSent() == 0 && nic.getBytesRecv() == 0) {
-                // nothing happens on this nic since boot
-                continue;
-            }
-            nics.add(new NicInfo(nic));
-        }
+        this.currentPid = extractCurrentPid();
```
ashangit: Add check on availibility of ss command otherwise exit fast innerCollect

Please add deployment of iproute package in test/src/main/docker/garmadon/Dockerfile

jvm-statistics/src/main/java/com/criteo/hadoop/garmadon/jvm/statistics/NetworkStatistics.java
```diff
@@ -2,86 +2,216 @@
...
+
+    private void collectNetworkStats(Map<Integer, ProcessInfo> processes, StatisticsSink sink) {
+        ProcessInfo processInfo = processes.get(currentPid);
+        updateSocketStats(processInfo);
+        for (ProcessInfo child : processInfo.children) {
+            updateSocketStats(child);
+        }
+        long recvBytes = 0;
+        long sentBytes = 0;
+        for (SocketStats ss : socketStats.values()) {
```
ashangit: Could we send diff to the event, more easier to manage with ES

PR #61, opened 2018-12-10
[HDFS EXPORTER] Decrease number of output files

gray-matter:
* Remove "path" event type (deprecated)
* Increase final files size
* Increase heartbeat period
* Set heartbeat period as non-multiple of expirer period

### DIFF
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/HdfsExporter.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/HdfsExporter.java
@@ -57,10 +57,11 @@ public class HdfsExporter {
             TMP_FILE_OPEN_RETRY_PERIOD);
 
     static {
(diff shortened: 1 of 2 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/HdfsExporter.java
```diff
@@ -57,10 +57,11 @@
             TMP_FILE_OPEN_RETRY_PERIOD);
 
     static {
-        DEFAULT_PROPERTIES_VALUE.put(MESSAGES_BEFORE_EXPIRING_WRITERS, 500000);
+        DEFAULT_PROPERTIES_VALUE.put(MESSAGES_BEFORE_EXPIRING_WRITERS, 3000000);
```
jpbempel: could use underscores
3_000_000 :)

PR #65, opened 2018-12-17
Add Presto module based on StandaloneModule

ashangit:
Presto use specific classloader for plugin with no parent.
This leads to garmadon class not being available for presto plugins
and also all libraries used by a plugin to not be loaded by parent classloader
so not being available for garmadon agent.
Using reflections on FileSystemTracer and ensuring that garmadon agent classes
are loaded from the App classloader and not the plugin one permits to use
trace hdfs call

### DIFF
--- a/agent/src/main/java/com/criteo/hadoop/garmadon/agent/modules/PrestoModule.java
+++ b/agent/src/main/java/com/criteo/hadoop/garmadon/agent/modules/PrestoModule.java
@@ -7,35 +7,18 @@ import com.criteo.hadoop.garmadon.agent.tracers.jvm.JVMStatisticsTracer;
 import com.criteo.hadoop.garmadon.agent.tracers.presto.PluginClassLoaderTracer;
 
 import java.lang.instrument.Instrumentation;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 
 public class PrestoModule implements GarmadonAgentModule {
 
     @Override
     public void setup(Instrumentation instrumentation, AsyncEventProcessor eventProcessor) {
-        ExecutorService executorService = Executors.newFixedThreadPool(5);
         // JVM/GC metrics/events
-        executorService.submit(() -> JVMStatisticsTracer.setup(
-                (timestamp, event) -> eventProcessor.offer(timestamp, StandaloneHeader.getInstance().getHeader(), event)));
+        JVMStatisticsTracer.setup((timestamp, event) ->
+                eventProcessor.offer(timestamp, StandaloneHeader.getInstance().getHeader(), event));
 
         // Byte code instrumentation
-        executorService.submit(() -> PluginClassLoaderTracer.setup(instrumentation));
-        executorService.submit(() -> FileSystemTracer.setup(instrumentation,
-                (timestamp, event) -> eventProcessor.offer(timestamp, StandaloneHeader.getInstance().getHeader(), event)));
-
-        executorService.shutdown();
-        // We wait 3 sec executor to instrument classes
-        // If all classes are still not instrumented after that time we let the JVM continue startup
-        // in order to not block the container for too long on agent initialization
-        // Currently we are seeing avg duration of 160 s on MapRed and 6500 s on SPARK
-        // so we consider 3s as a reasonable duration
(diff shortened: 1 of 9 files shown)


agent/src/main/java/com/criteo/hadoop/garmadon/agent/modules/PrestoModule.java
```diff
@@ -0,0 +1,41 @@
...
+        executorService.submit(() -> PluginClassLoaderTracer.setup(instrumentation));
+        executorService.submit(() -> FileSystemTracer.setup(instrumentation,
+                (timestamp, event) -> eventProcessor.offer(timestamp, StandaloneHeader.getInstance().getHeader(), event)));
+
+        executorService.shutdown();
+        // We wait 3 sec executor to instrument classes
+        // If all classes are still not instrumented after that time we let the JVM continue startup
+        // in order to not block the container for too long on agent initialization
+        // Currently we are seeing avg duration of 160 s on MapRed and 6500 s on SPARK
+        // so we consider 3s as a reasonable duration
```
piyushnarang: slightly confused here. If we see an avg duration on map red and spark which is much much larger, why is 3s reasonable? Or am I missing something?

pom.xml
```diff
@@ -91,6 +91,7 @@
         <guava.version>20.0</guava.version>
         <parquet-protobuf.version>1.8.1</parquet-protobuf.version>
         <protobuf-dynamic.version>0.9.4</protobuf-dynamic.version>
+        <presto.version>0.206</presto.version>
```
piyushnarang: does this mean on every presto release we need to also bump and rollout Garmadon?

agent/src/main/java/com/criteo/hadoop/garmadon/agent/modules/PrestoModule.java
ashangit: My bad I should remove this comment from this module and the timeout
It is required for ContainerModule because we don't want to stuck container too long initialising the agent
On long running applications like nodemanagers, presto... delaying a few startup waiting complete initialisation of the agent is not an issue

pom.xml
ashangit: No unless presto change is classloading mechanism + classes associated to it.

agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/hdfs/FileSystemTracer.java
```diff
@@ -77,9 +72,10 @@ protected Implementation newImplementation() {
         public static Object intercept(
                 @SuperCall Callable<?> zuper,
                 @This Object o,
-                @Argument(0) Path dst) throws Exception {
-            Object uri = ((DistributedFileSystem) o).getUri();
-            return executeMethod(zuper, uri.toString(), null, dst.toString(), FsAction.DELETE.name());
+                @Argument(0) Object dst) throws Exception {
+            Class distributedFileSystem = o.getClass().getClassLoader().loadClass("org.apache.hadoop.hdfs.DistributedFileSystem");
+            Method getUri = distributedFileSystem.getMethod("getUri");
```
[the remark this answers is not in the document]
ashangit: I will check if I can improve this
But I don't see how I can precompute field as I need to use the classloader on which the objects being intercept has been created. It is not the same than the one for garmadon agent classes and not accessible from it on presto use cases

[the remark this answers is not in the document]
Willymontaz: We need to load the class from the classloader of arg o. If not we perform illegal calls later in the code. Is that costly to invoke o.getClass().getClassLoader().loadClass() once the class is loaded. Would we gain from maintaining a Map<ClassLoader, Class> ?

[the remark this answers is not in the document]
jpbempel: @ashangit ok I missed that part. the class is always the same, but ClassLoader can be different.
@Willymontaz loading a class already loaded is not costly, this is obtaining an instance of the method that is more
so you can keep Map<ClassLoader, Method> instead.
But put this only if there is really an issue on perf here.

PR #72, opened 2018-12-31
Update template at reader startup

ashangit:


test/src/main/docker/garmadon/scripts/entrypoint.sh
```diff
@@ -8,6 +8,10 @@ function check_resourcemanager_up {
     until curl http://resourcemanager:8088/; do sleep 1; done
 }
 
+function check_es_up {
```
Willymontaz: This change should be in a separate review

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -176,10 +210,15 @@ private static BulkProcessor setUpBulkProcessor(String esHost, int esPort, Strin
         //setup es client
         RestHighLevelClient esClient = new RestHighLevelClient(restClientBuilder);
 
+        putGarmadonTemplate(esClient, elasticsearch);
+
         Sniffer sniffer = Sniffer.builder(esClient.getLowLevelClient()).build();
         sniffOnFailureListener.setSniffer(sniffer);
 
-        return BulkProcessor.builder(esClient::bulkAsync, new ElasticSearchListener())
+        BiConsumer<BulkRequest, ActionListener<BulkResponse>> bulkConsumer =
```
Willymontaz: what is this change for ?

test/src/main/docker/garmadon/scripts/entrypoint.sh
ashangit: No I need it now since I load template from ES reader

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
ashangit: Due to bump of ES to 6.5

Willymontaz: Was just curious since we use RequestOptions.DEFAULT. I'm ok with the way it is done here :)

test/src/main/docker/garmadon/scripts/entrypoint.sh
Willymontaz: Ok got it, that was for the block_until_website_available that you removed

PR #74, opened 2019-01-02
Add back normalization on FS_EVENT removed when migrating to ProtoConcatenator concatToMap function

ashangit:
Due to this bug uri can be different for same nameservice access and path can contain hdfs namenservice

### DIFF
--- a/readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
+++ b/readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
@@ -98,7 +98,7 @@ public final class ElasticSearchReader {
 
     void writeToES(GarmadonMessage msg) {
         String msgType = GarmadonSerialization.getTypeName(msg.getType());
-        if ("JVMSTATS_EVENT".equals(msgType)) {
+        if (GarmadonSerialization.TypeMarker.JVMSTATS_EVENT == msg.getType()) {
             Map<String, Object> jsonMap = ProtoConcatenator.concatToMap(Arrays.asList(msg.getHeader()), true);
 
             HashMap<String, Map<String, Object>> eventMaps = new HashMap<>();
--- a/readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
+++ b/readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
@@ -113,7 +113,7 @@ public final class ElasticSearchReader {
             eventMap.put("event_type", msgType);
 
             // Specific normalization for FS_EVENT
-            if ("FS_EVENT".equals(msgType)) {
+            if (GarmadonSerialization.TypeMarker.FS_EVENT == msg.getType()) {
                 String uri = (String) eventMap.get("uri");
                 eventMap.computeIfPresent("src_path", (k, v) -> ((String) v).replace(uri, ""));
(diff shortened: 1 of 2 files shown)


readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -111,6 +111,15 @@ void writeToES(GarmadonMessage msg) {
         } else {
             Map<String, Object> eventMap = ProtoConcatenator.concatToMap(Arrays.asList(msg.getHeader(), msg.getBody()), true);
             eventMap.put("event_type", msgType);
+
+            // Specific normalization for FS_EVENT
+            if ("FS_EVENT".equals(msgType)) {
+                String uri = (String) eventMap.get("uri");
+                eventMap.computeIfPresent("src_path", (k, v) -> ((String) v).replace(uri, ""));
```
gray-matter: I'm guessing we disregard cases where uri might be in the middle of these strings (just checking)

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -111,6 +111,15 @@ void writeToES(GarmadonMessage msg) {
...
+                eventMap.computeIfPresent("dst_path", (k, v) -> ((String) v).replace(uri, ""));
+                eventMap.computeIfPresent("uri", (k, v) -> UriHelper.getUniformizedUri((String) v));
```
gray-matter: This means that no matter where the file is coming from, we consider its prefix to be that of the current cluster? If we start copying files between different clusters, wouldn't that be an issue?

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
ashangit: "hdfs://<nameservice>" or "viewfs://<namservice" should not, I think we can disregard

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -111,6 +111,15 @@ void writeToES(GarmadonMessage msg) {
         } else {
             Map<String, Object> eventMap = ProtoConcatenator.concatToMap(Arrays.asList(msg.getHeader(), msg.getBody()), true);
             eventMap.put("event_type", msgType);
+
+            // Specific normalization for FS_EVENT
+            if ("FS_EVENT".equals(msgType)) {
```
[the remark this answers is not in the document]
ashangit: ?

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
gray-matter: OK, sorry I misread the code. This is more like a "base URI"/"namespace URI" than an actual full URI

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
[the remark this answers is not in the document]
gray-matter: Some constant holding "FS_EVENT". Just curious whether this is used somewhere else, in which case a constant could be useful (but not mandatory).

PR #76, opened 2019-01-07
Fix Kafka Prometheus metrics

gray-matter:
Metrics is always found using consumer-1, not the hostname


gray-matter:
Fixing the consumer id instead, PR incoming

PR #81, opened 2019-01-14
[HDFS exporter] Batch HDFS calls to get initial offset

gray-matter:
Before this change, nb_partitions x nb_events calls were made to HDFS.
Since these calls can sometimes last for a few 10s of seconds, Kafka could
drop (was on prod) consumers and loop-reassign partitions

This change decreases calls to nb_events calls. This could be changed to only
one but would require further changes. Given the current amount of events, this
should work fine for the time being.


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/kafka/OffsetResetter.java
```diff
@@ -52,41 +50,54 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
 
     @Override
     public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
-         for (TopicPartition partition : partitions) {
-            long startingOffset = UNKNOWN_OFFSET;
-            List<Long> offsets = new ArrayList<>(writers.size());
+        Map<Integer, Collection<Long>> offsetsPerPartition = new HashMap<>();
```
Willymontaz: I think it could be good to explain here that the way the code is shapped is to collect offsets in a fast manner to prevent kafka from considering the consumer as dead.

PR #84, opened 2019-01-18
Retreive real hdfs user performing hdfs call

### LINKED ISSUE #83: For hdfs event user reported should be the one from DFS not the user running the process
Currently the user reported as the user performing the hdfs call is the one running the java process.
This is wrong especially for applications like hive/presto.
We need to find a way to send another user field from dfsclient or other.

ashangit:
On application like presto/hive the user performing the hdfs call is not the user running the JVM instance
This commit will provide real hdfs user information except for ADD_BLOCK actions as we don't have any simple mechanism
to retreive it

Fixes #83


agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/hdfs/FileSystemTracer.java
```diff
@@ -311,23 +297,34 @@ public static Object intercept(
             Method getServerAddress = getMethod(classLoader, "org.apache.hadoop.ipc.RPC", "getServerAddress", Object.class);
             InetSocketAddress inetSocketAddress = (InetSocketAddress) getServerAddress.invoke(o, rpcProxy);
             return executeMethod(zuper, "hdfs://" + inetSocketAddress.getHostString() + ":" + inetSocketAddress.getPort(),
-                    null, dst, FsAction.ADD_BLOCK.name());
+                    null, dst, FsAction.ADD_BLOCK.name(), null);
         }
     }
 
-    private static Object executeMethod(@SuperCall Callable<?> zuper, String uri, String src, String dst, String fsAction) throws Exception {
+    private static Object callDistributedFileSystem(@SuperCall Callable<?> zuper, @This Object o, String src, String dst, String fsAction) throws Exception {
```
gray-matter: Any way to test this? This seems very error-prone and could benefit from automated tests

agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/hdfs/FileSystemTracer.java
```diff
@@ -311,23 +297,34 @@ public static Object intercept(
...
+        ClassLoader classLoader = o.getClass().getClassLoader();
+        Field dfsField = getField(classLoader, "org.apache.hadoop.hdfs.DistributedFileSystem", "dfs");
```
gray-matter: Any reason for not using something like .class.getCanonicalName() instead of strings? (I'm guessing maybe because we don't want to depend on these packages)

ashangit: Please look at #65

agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/hdfs/FileSystemTracer.java
ashangit: done

agent/src/test/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/hdfs/FileSystemTracerTest.java
```diff
@@ -96,174 +103,127 @@ public static void setUpClass() throws IOException, NoSuchFieldException, Illega
...
-
     }
 
-    @Test
-    @AgentAttachmentRule.Enforce
-    public void FileSystemTracer_should_attach_to_write() throws ClassNotFoundException, NoSuchMethodException, IllegalAccessException, InvocationTargetException, URISyntaxException {
-        assertThat(ByteBuddyAgent.install(), instanceOf(Instrumentation.class));
-
-        ClassFileTransformer classFileTransformer = new FileSystemTracer.WriteTracer().installOnByteBuddyAgent();
+    private void checkEvent(String action, Path path) {
```
gray-matter: Shouldn't this check the user too? (purpose of this PR)

agent/src/test/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/hdfs/FileSystemTracerTest.java
```diff
@@ -150,11 +169,13 @@ private void checkEvent(String action, Path path) {
         assertEquals(action, eventTmp.getAction());
         assertEquals(hdfsURI, eventTmp.getUri());
         assertEquals(path.toString(), eventTmp.getDstPath());
+
+        assertEquals(hdfs_user, eventTmp.getHdfsUser());
```
gray-matter: OPT: Should read hdfsUser instead

PR #89, opened 2019-01-22
[reader] serialize timestamp with along header and body

mchataigner:
Currently ES reader is sending timestamp as a special field but hadoop
reader is using parquet and was not sending it.

This review adds a field `timestamp` when merging header and body to a
single protobuf message.

This field comes from GarmadonMessage.getTimestamp which is the
timestamp in millisecond of the event.


readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -117,7 +117,7 @@ void writeToES(GarmadonMessage msg) {
                 addEventToBulkProcessor(eventMap, msg.getTimestamp(), msg.getCommittableOffset());
             }
         } else {
-            Map<String, Object> eventMap = ProtoConcatenator.concatToMap(Arrays.asList(msg.getHeader(), msg.getBody()), true);
+            Map<String, Object> eventMap = ProtoConcatenator.concatToMap(msg.getTimestamp(), Arrays.asList(msg.getHeader(), msg.getBody()), true);
```
ashangit: please remove  eventMap.put("timestamp", timestamp); in private void addEventToBulkProcessor(Map<String, Object> eventMap, long timestamp, CommittableOffset committableOffset) {

ashangit: And remove msg.getTimestamp in             addEventToBulkProcessor(eventMap, msg.getTimestamp(), msg.getCommittableOffset());

mchataigner: Done

PR #93, opened 2019-01-24
[HDFS exporter] Kill the app when failing to close output file

gray-matter:
This is usually indicative of an unrecoverable situation.


mchataigner:
Could you update Garmadon.stopReading and return cf every time?

PR #98, opened 2019-02-01
Add tags mechanism on agent

### LINKED ISSUE #97: Improve tags on agent
Currently tags are hardcoded on agent header.
We need to improve this so we can push an array of tags with tags define by user.
For ex. providing a garmadon.tags java properties on which user can set the presto server name and other tags that then could help filter event in ES

ashangit:
Closes #97


Willymontaz:
I propose to force Uppercase (for consistency with other tags), alpha-numeric and '-', '_', '.' to avoid bad surprises

PR #110, opened 2019-02-19
Support kafka cluster migration in reader-hdfs

mchataigner:
Add kafka cluster name in file naming convention.
Be backward compatible.
if kafka.cluster is defined use it for file pattern
else use the old pattern

### DIFF
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
@@ -27,6 +27,7 @@ public class HdfsOffsetComputer implements OffsetComputer {
     private final FileSystem fs;
     private final Path basePath;
     private final String kafkaCluster;
+    private final String fileRenamePattern;
 
     public HdfsOffsetComputer(FileSystem fs, Path basePath) {
         this(fs, basePath, null);
@@ -43,8 +44,10 @@ public class HdfsOffsetComputer implements OffsetComputer {
         this.kafkaCluster = kafkaCluster;
         if (kafkaCluster == null) {
             this.offsetFilePatternGenerator = Pattern.compile("^(\\d+)\\.(\\d+)$");
+            this.fileRenamePattern = "%s/%d.%d";
         } else {
             this.offsetFilePatternGenerator = Pattern.compile(String.format("^(\\d+)\\.cluster=%s\\.(\\d+)$", kafkaCluster));
+            this.fileRenamePattern = String.format("%%s/%%d.cluster=%s.%%d", kafkaCluster);
         }
     }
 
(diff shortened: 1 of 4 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -8,82 +8,78 @@
...
      * @param basePath                      Root directory to look in for offsets filenames
+     * @param kafkaCluster                  Name of corresponding kafka cluster
      */
-    public HdfsOffsetComputer(FileSystem fs, Path basePath) {
+    public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCluster) {
         this.fs = fs;
         this.basePath = basePath;
-        this.offsetFilePatternGenerator = partitionId -> Pattern.compile(String.format("^%d.(\\d+)$", partitionId));
+        this.kafkaCluster = kafkaCluster;
+        if (kafkaCluster == null) {
```
Willymontaz: Can we avoid using null and create a third constructor that injects offsetFilePatternGenerator ?

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -8,82 +8,78 @@
...
-        return highestOffset;
+        return result;
     }
 
     @Override
     public String computePath(LocalDateTime time, Offset offset) {
-        return String.format("%s/%d.%d", time.format(DateTimeFormatter.ISO_DATE), offset.getPartition(),
-                offset.getOffset());
+        if (kafkaCluster == null) {
+            return String.format("%s/%d.%d", time.format(DateTimeFormatter.ISO_DATE), offset.getPartition(),
```
Willymontaz: can we rely on injection at construction and avoid null checking ?

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
mchataigner: I would avoid putting pattern outside of this class because previous file pattern is tied to file rename pattern.
kafkaCluster may be null because it comes from kafkaConfiguration which is parsed by jackson-databind, hence may be null if not set in the config.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
mchataigner: done.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
Willymontaz: ok

PR #111, opened 2019-03-04
[HDFS exporter] Only poll last 2 days of logs when computing offsets

gray-matter:
This used to be listing _all_ files recursively. This is a bit overkill and
resource-consuming. In some very specific cases, this could still cause events
duplication.


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/HdfsExporter.java
```diff
@@ -133,7 +133,7 @@ public static void main(String[] args) throws IOException {
             final Class<? extends Message> clazz = out.getValue().getValue();
             final Function<LocalDateTime, ExpiringConsumer<Message>> consumerBuilder;
             final Path finalEventDir = new Path(finalHdfsDir, eventName);
-            final OffsetComputer offsetComputer = new HdfsOffsetComputer(fs, finalEventDir, config.getKafka().getCluster());
+            final OffsetComputer offsetComputer = new HdfsOffsetComputer(fs, finalEventDir, config.getKafka().getCluster(), 2);
```
ashangit: can we make this configurable?

gray-matter: Done

PR #114, opened 2019-03-05
fix javadoc error

mchataigner:


ashangit:
plz also fix HdfsExporter @throws IOException line 81

PR #116, opened 2019-03-11
[HDFS EXPORTER] Output the exception when failing to close a consumer

gray-matter:
Next step would be to fail the app when failing too often. Otherwise a given even type could be
lagging behind by too much


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/PartitionedWriter.java
```diff
@@ -269,7 +269,7 @@ private boolean tryExpireConsumer(ExpiringConsumer<MESSAGE_KIND> consumer) {
                 consumer.close();
                 return true;
             } catch (IOException e) {
-                LOGGER.error("Couldn't close writer for {} ({}/{})", eventName, retry, maxAttempts, e);
+                LOGGER.error("Couldn't close writer for {} ({}/{}): {}", eventName, retry, maxAttempts, e.getMessage());
```
mchataigner: use String.format and pass the exception as second parameter to LOGGER.error to keep stack trace

PR #117, opened 2019-03-13
Support Flink framework on Garmadon

ashangit:
Currently only compatible with ES and HDFS reader
Required to add this config on jobs
  -yD metrics.reporters=garmadon \
  -yD metrics.reporter.garmadon.class=com.criteo.hadoop.garmadon.flink.GarmadonFlinkReporter

This is still an on going PR but any feedback at this step will be welcome.
Still missing ES dashboards + some testing with some real flink jobs


ashangit:
@gray-matter @mchataigner @Willymontaz plz review

PR #118, opened 2019-03-15
Remove temporary hdfs folder at reader shutdown

ashangit:
Currently folders in hdfs temp folder based on UUID are never deleted leading
to lots of files/folders beeing kept on hdfs
On normal run we must ensure that those folders are reclaimed at shutdown

### DIFF
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/HdfsExporter.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/HdfsExporter.java
@@ -67,6 +67,8 @@ public class HdfsExporter {
     private static Duration tmpFileOpenRetryPeriod;
     private static int sizeBeforeFlushingTmp;
 
+    private static FileSystem fs;
+
     protected HdfsExporter() {
         throw new UnsupportedOperationException();
     }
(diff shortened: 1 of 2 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/HdfsExporter.java
```diff
@@ -88,15 +90,23 @@ public static void main(String[] args) throws IOException {
...
             exit(1);
         }
 
+        // Add Shutdown Hook to close FS and delete temp folders
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            try {
+                fs.close();
+            } catch (IOException e) {
+                LOGGER.error("Could not close FS", e);
+                exit(1);
```
mchataigner: why do you need this shutdown hook? it will be closed anyway by the shutdown of the JVM or is it required by the deleteOnExit?

ashangit: How could it be closed during the shutdown of the JVM?
To process deletion of deleteOnExit path fs object should call the processDeleteOnExit method which is called on the close method.
This is why I'm using this shutdown hook. Also it ensure that the close "action" will be closed in almost all cases, only JVM crash and SIGKILL will leads to this action not being called.

mchataigner: there is already a shutdown hook for filesystems in FileSystem.Cache.getInternal

ashangit: Good catch

PR #119, opened 2019-03-19
First MR to implement merge of parquet file on hdfs reader

ashangit:
Write offset for each messages on each parquet table
Add index on filename

Idea is then to be able from index to get last written files per day and then from parquet metadata to get max offset.

PR #120, opened 2019-03-19
Second MR to implement merge of parquet file on hdfs reader

ashangit:
Do not merge before #119 
2 first commits are part of #119 but are needed for this one

### DIFF
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
@@ -2,6 +2,9 @@ package com.criteo.hadoop.garmadon.hdfs.offset;
 
 import com.criteo.hadoop.garmadon.reader.Offset;
 import org.apache.hadoop.fs.*;
+import org.apache.parquet.column.statistics.LongStatistics;
+import org.apache.parquet.hadoop.ParquetFileReader;
+import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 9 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +48,87 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+
+                    })
+                    .reduce(NO_OFFSET, (max1, max2) -> max1 > max2 ? max1 : max2);
+
+            maxOffsets.add(maxValue);
         }
 
-        return "{" + String.join(",", dirnames) + "}";
+        // Get max offset from all files for a partition
+        return maxOffsets.stream().reduce(NO_OFFSET, (max1, max2) -> max1 > max2 ? max1 : max2);
```
mchataigner: return maxOffsets.stream().mapToLong(Long::longValue).max().orElse(NO_OFFSET);

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
```diff
@@ -74,26 +79,76 @@ public Path close() throws IOException {
...
-
         final Path topicGlobPath = new Path(finalHdfsDir, fileNamer.computeTopicGlob(dayStartTime, latestOffset));
-        final Optional<Long> lastIndex = Arrays.stream(fs.globStatus(topicGlobPath))
-                .map(file -> fileNamer.getIndex(file.getPath().getName()))
-                .reduce((index1, index2) -> index1 > index2 ? index1 : index2);
 
-        final Path finalPath = new Path(finalHdfsDir, fileNamer.computePath(dayStartTime, lastIndex.orElse(0L) + 1, latestOffset));
+        final Optional<Path> lastAvailableFinalPath = Arrays.stream(fs.globStatus(topicGlobPath))
+                .map(FileStatus::getPath)
+                .reduce((path1, path2) -> fileNamer.getIndex(path1.getName()) > fileNamer.getIndex(path2.getName()) ? path1 : path2);
```
mchataigner: .max(Comparator.comparingInt(path -> fileNamer.getIndex(path.getName())))

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
```diff
@@ -74,26 +79,76 @@ public Path close() throws IOException {
...
+            moveToFinalPath(mergedTempFile, lastAvailableFinalPath);
+            try {
+                fs.delete(temporaryHdfsPath, false);
+                // This file is in a temp folder that should be deleted at exit so we should not throw exception here
+            } catch (IOException ignored) {
+            }
+        }
+    }
+
+    private Boolean checkSchemaEquality(MessageType schema) throws IOException {
```
mchataigner: boolean

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
```diff
@@ -74,26 +79,76 @@ public Path close() throws IOException {
...
+        MessageType schema2 = ParquetFileReader.open(fs.getConf(), temporaryHdfsPath).getFileMetaData().getSchema();
+
+        return schema.getColumns().size() == schema2.getColumns().size() &&
```
mchataigner: why not schema.equals(schema2) ?
or schema.checkContains(schema2) || schema2.checkContains(schema) ?

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +48,87 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+                        LOGGER.info("Couldn't deviate a valid offset from '{}'", fileName);
                     }
-                } catch (NumberFormatException e) {
-                    LOGGER.info("Couldn't deviate a valid offset from '{}'", fileName);
                 }
             }
         }
 
+        // Get last offset
+        for (int partitionId : resultFile.keySet()) {
```
Willymontaz: could we process offsets right away and avoid the use of the intermediate hashmap ?

Otherwise, if you want to keep the hashmap, you can iterate on hashmap entrySet instead of using keySet + then getting again from the map with the key.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +48,87 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
-    private String computeDirNamesPattern(LocalDateTime today, int backlogDays) {
-        List<String> dirnames = new ArrayList<>(backlogDays);
-
-        for (int i = 0; i < backlogDays; ++i) {
-            LocalDateTime day = today.minusDays(i);
-
-            dirnames.add(computeDirName(day));
+    protected Long getMaxOffset(Map<String, FinalEventPartitionFile> dateFinalEventPartitionFile) throws IOException {
+        List<Long> maxOffsets = new ArrayList<>();
+        for (FinalEventPartitionFile finalEventPartitionFile : dateFinalEventPartitionFile.values()) {
```
Willymontaz: you could also stream dateFinalEventPartitionFile.values() instead of iterating. Streams can be parallelised, is it something we could benefit since we might have many files to read ?

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
ashangit: checkContains -> return void so for sure I can't use it
Will check for schema.equals(schema2)

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +48,87 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+                                (!resultFile.containsKey(partitionId) || !resultFile.get(partitionId).containsKey(listedDay) ||
+                                        index > resultFile.get(partitionId).get(listedDay).getIndex())) {
+                            HashMap<String, FinalEventPartitionFile> dateFinalEventPartitionFile =
+                                    (HashMap<String, FinalEventPartitionFile>) resultFile.computeIfAbsent(partitionId, s -> new HashMap<>());
+                            dateFinalEventPartitionFile.put(listedDay,
+                                    new FinalEventPartitionFile(index, status.getPath()));
+                            resultFile.put(partitionId, dateFinalEventPartitionFile);
+                        }
+                    } catch (NumberFormatException e) {
+                        LOGGER.info("Couldn't deviate a valid offset from '{}'", fileName);
```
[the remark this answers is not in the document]
ashangit: It is for this parsing: Long index = Long.parseLong(matcher.group("index"));

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
ashangit: I use hashmap to only read last index file per day and not to read offset on each files
So seems for me necessary to use it. Let me know if you think about something else.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
ashangit: We could have 2 files to read in // with current setup so yes lets use stream

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +50,107 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+                            .findFirst();
+
+                        if (columnChunkMetaData1.isPresent()) {
+                            return ((LongStatistics) columnChunkMetaData1.get().getStatistics()).genericGetMax();
+                        } else {
+                            return NO_OFFSET;
+                        }
+
+                    })
+                    .reduce(NO_OFFSET, (max1, max2) -> max1 > max2 ? max1 : max2);
```
mchataigner: .max().orElse(NO_OFFSET)

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +50,107 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+    private boolean isPartitionDayNotAlreadyInHashPartitionDay(int partitionId, String listedDay, Map<Integer,
+        Map<String, FinalEventPartitionFile>> resultFile) {
+        return !(resultFile.containsKey(partitionId) && resultFile.get(partitionId).containsKey(listedDay));
+    }
 
-            dirnames.add(computeDirName(day));
-        }
+    protected Long getMaxOffset(Map<String, FinalEventPartitionFile> dateFinalEventPartitionFile) {
+        List<Long> maxOffsets = new ArrayList<>();
+        dateFinalEventPartitionFile.values().stream()
```
Willymontaz: regarding my last comment around stream and parallelism. It is usually more suited to big list processing with ForkJoin pool but we could give it a try. Just few things that needs to be done to allow this:
- get parallelStream instead of stream
- do not put side effect in the lambdas processing the stream. Here it is the fact that we add maxValue to maxOffsets from within the lambda. If we go parallel, there will be synchronisation issues. The proper way to handle this is to use map then collect instead of foreach to create the maxOffsets list.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +50,107 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+                        if (columnChunkMetaData1.isPresent()) {
+                            return ((LongStatistics) columnChunkMetaData1.get().getStatistics()).genericGetMax();
+                        } else {
+                            return NO_OFFSET;
+                        }
+
+                    })
+                    .reduce(NO_OFFSET, (max1, max2) -> max1 > max2 ? max1 : max2);
+
+                maxOffsets.add(maxValue);
```
Willymontaz: see previous comment, would be nice to use map then collect instead of mutating a list created outside of the lambda

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +50,107 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+                }
+
+                // Get max value from all blocks
+                long maxValue = pFR.getFooter().getBlocks().stream()
+                    .map(b -> {
+                        Optional<ColumnChunkMetaData> columnChunkMetaData1 = b.getColumns().stream()
+                            .filter(column -> Arrays.stream(column.getPath().toArray()).allMatch(path -> path.equals("kafka_offset")))
+                            .findFirst();
+
+                        if (columnChunkMetaData1.isPresent()) {
```
Willymontaz: columnChunkMetaData1
                        .map(ColumnChunkMetaData::getStatistics)
                        .map(Statistics::genericGetMax)
                        .map(Long.class::cast)
                        .orElse(NO_OFFSET)

So actually you can chain everything directly from b.getColumns().stream() etc....

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +50,107 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
-            dirnames.add(computeDirName(day));
-        }
+    protected Long getMaxOffset(Map<String, FinalEventPartitionFile> dateFinalEventPartitionFile) {
+        List<Long> maxOffsets = new ArrayList<>();
+        dateFinalEventPartitionFile.values().stream()
+            .forEach(finalEventPartitionFile -> {
+                // Read parquet file
+                ParquetFileReader pFR;
+                try {
+                    pFR = ParquetFileReader.open(fs.getConf(), finalEventPartitionFile.getFilePath());
```
Willymontaz: it looks like pFR is never closed. It is safer to use try-with-resource

 try (ParquetFileReader pFR = ParquetFileReader.open(fs.getConf(), finalEventPartitionFile.getFilePath())) {
   //stream processing
} catch [...]

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HdfsOffsetComputer.java
```diff
@@ -45,51 +50,107 @@ public HdfsOffsetComputer(FileSystem fs, Path basePath, @Nullable String kafkaCl
...
+                        }
+
+                    })
+                    .reduce(NO_OFFSET, (max1, max2) -> max1 > max2 ? max1 : max2);
+
+                maxOffsets.add(maxValue);
+            });
+
+        // Get max offset from all files for a partition
+        return maxOffsets.stream().mapToLong(Long::longValue).max().orElse(NO_OFFSET);
```
Willymontaz: Instead of populating a list and then stream on it to find max, you can use directly the stream dateFinalEventPartitionFile.values().stream() at line 122

PR #121, opened 2019-03-20
Add grafana dashboard for Nodemanagers JVM

Willymontaz:


ashangit:
Could we only have one dashboards for nodemanagers/resourcemanager/standalone JVM?

Willymontaz:
I will try to make on dashboard as @ashangit suggested. If it is to difficult to do so, I'll merge with a separate dashboard

PR #128, opened 2019-04-05
Add support of prestosql fork to PrestoModule

rclaude:


agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/presto/PluginClassLoaderTracer.java
```diff
@@ -46,15 +50,35 @@ protected Implementation newImplementation() {
...
+                    success = true;
+                } catch (Exception e) {
+                    exception = e;
+                }
+            }
 
-            spiClassLoaderField.set(o, ImmutableList.copyOf(tmpspiPackages));
+            if (!success) {
+                throw exception;
+            }
```
ashangit: What about just using o.getClass().getName() instead of pluginClassLoaderClassName to get the appropriate plugin name

`Class pluginClassLoaderZz = o.getClass().getClassLoader().loadClass(o.getClass().getName());
            Field spiClassLoaderField = pluginClassLoaderZz.getDeclaredField("spiPackages");
            spiClassLoaderField.setAccessible(true);

            ImmutableList<String> spiPackages = (ImmutableList<String>) spiClassLoaderField.get(o);
            List<String> tmpspiPackages = new ArrayList<>(spiPackages);
            tmpspiPackages.add("com.criteo.hadoop.garmadon.agent.");

            spiClassLoaderField.set(o, ImmutableList.copyOf(tmpspiPackages));

ashangit:
Also add a test with new io.prestosql.server.PluginClassLoader so we can spot any change in presto that can affect this Tracer

PR #131, opened 2019-04-16
HDFS reader indicates when it's done with a given day

gray-matter:
Write a "done" file for D-1 when D + 2h is handled


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
```diff
@@ -0,0 +1,68 @@
...
+        Path path = checkpointPathGenerator.apply(when);
+
+        try {
+            if (!checkpointed(path)) {
+                checkpoint(path);
+                checkpointsCache.put(path, true);
+
+                return true;
+            } else {
+                checkpointsCache.put(path, true);
```
Willymontaz: This line is useless

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.function.Function;
+
+/**
+ * Write checkpoints on the filesystem.
+ */
+public class FsBasedCheckpointer implements Checkpointer {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FsBasedCheckpointer.class.getName());
+
+    private final FileSystem fs;
+    private final Map<Path, Boolean> checkpointsCache;
```
Willymontaz: The cache is never cleaned. Even if it is not containing a lot of data I don't like the idea that eventually we fill memory with useless data until the JVM dies.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private void checkpoint(Path path) throws IOException {
+        FSDataOutputStream os = fs.create(path);
+
+        os.close();
+    }
+
+    private boolean checkpointed(Path path) throws IOException {
+        if (!checkpointsCache.containsKey(path)) {
```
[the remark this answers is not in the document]
gray-matter: I can't because of the IOException fs.exists throws and the second argument of computeIfAbsent cannot throw

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
gray-matter: Removed

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/DelayedDailyPathComputer.java
```diff
@@ -0,0 +1,24 @@
+package com.criteo.hadoop.garmadon.hdfs.writer;
+
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.Temporal;
+import java.time.temporal.TemporalAmount;
+import java.util.function.Function;
+
+public class DelayedDailyPathComputer implements Function<Temporal, String> {
```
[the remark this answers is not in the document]
gray-matter: Because the point of this class is to compute a String path from a Temporal argument? :-)

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
gray-matter: We're talking about nb_events x uptime days, it would take a very long time to be of a significative size...while I agree with you on the principle, I would qualify this as premature optimization

gray-matter:
> With this implementation, a reader instance can checkpoint a day even if we are very late on some partitions and still have a lot of messages to output for that day.

True that, will fix!

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/DelayedDailyPathComputer.java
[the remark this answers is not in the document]
Willymontaz: But it does not need to implement this interface to do so

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
Willymontaz: When you will take in account partitions, that will be more data. 
Cleaning the cache is not about optimisation, it is more about managing memory properly and avoid obvious leaks. Is there such thing as "premature correct memory management" ?

Willymontaz:
Also consider that you need to take in account the commit phase rather than trying to checkpoint at every message. Even if you take partitions into account, if you checkpoint right away, there is a great chance for the events not to be committed on HDFS yet.
Checkpointing at the commit phase won't need a cache anymore to avoid huge accesses to HDFS.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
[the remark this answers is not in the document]
ashangit: I think you can throw a RuntimeException if needed

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
ashangit: @gray-matter you can use guva cache for this (already loaded by hadoop dependencies)
I don't se this as a premature optimization but as a bug
Design which rely on probable restart of instances before the OOM issue doesn't looks good to me

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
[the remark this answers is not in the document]
gray-matter: But then, what's the point? Using computeIfAbsent here would be mostly cosmetic, this is as explicit

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/DelayedDailyPathComputer.java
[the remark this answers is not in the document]
gray-matter: I like to make it clear what the purpose of a class is when it looks like a function, a producer or something similar. I don't think it hurts, but yes, it's not purely useful :smile:

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/FsBasedCheckpointer.java
gray-matter: Done

ashangit:
What's happen if a reader crash with a partition for which there are no more messages but it hasn't reach the 2H meaning no checkpoint have been created?

gray-matter:
> What's happen if a reader crash with a partition for which there are no more messages but it hasn't reach the 2H meaning no checkpoint have been created?

* If a reader crashes, it would resume from where it left the processing.
* If there's no message for a partition, then there's nothing to process, so the "done" file is useless.
* Based on how the DelayedPathComputer works, committing anything between D + 2h until D + 26h would create a "done" file for D, which seems far enough wrt the other delays.

Does that answer your question?

ashangit:
LGTM

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/PartitionedWriter.java
```diff
@@ -190,13 +203,18 @@ private void possiblyCloseConsumers(Predicate<ExpiringConsumer> shouldClose) {
...
+                    final LocalDateTime day = entry.getKey();
 
                     if (shouldClose.test(consumer)) {
                         if (tryExpireConsumer(consumer)) {
                             final Counter.Child filesCommitted = PrometheusMetrics.buildCounterChild(
                                 PrometheusMetrics.FILES_COMMITTED, eventName);
 
                             filesCommitted.inc();
+
+                            checkpointer.tryCheckpoint(partitionId, latestMessageTimeForPartitionAndDay.get(
```
Willymontaz: if checkpointing fails due to an exception, we want to continue and try to checkpoint at the next commit right ?
If so we should protect with try/catch.

Also, I would like to add a prometheus gauge to create an alert on checkpointing not performed for a partition/day

gray-matter: Done

PR #132, opened 2019-04-19
Enrich events related to application running on the cluster on ES reader

ashangit:
In order to be able to filter vcore/memory metrics per application_name, users, frameworks or component
we have to enrich those events with informations from AppEvent and events from containers (Doing joins on elasticsearch is not possible/efficient).
As component is only a notion from the framework we can only get this information from containers class parsing.
This information could also be set to UNKNOWN if we don't know the class which make impossible to display AM metrics or workers metrics.
In order to do so we are retreiving am container id from RM and compare it from current ccontainer id and update the component to APP_MASTER otr APP_SLAVE.


readers/elasticsearch/src/test/java/com/criteo/hadoop/garmadon/elasticsearch/cache/ElasticSearchCacheManagerTest.java
```diff
@@ -0,0 +1,222 @@
...
+        assertEquals(framework, eventMap.get("framework"));
+        assertEquals(username, eventMap.get("username"));
+        assertEquals(yarnTags, eventMap.get("yarn_tags"));
+        assertEquals(component, eventMap.get("component"));
+    }
+
+    @Test
+    public void enrich_unknown_component_event_with_app_in_the_cache() {
+        String unknownContainerId = containerId + "_2";
+        eventMap.put("container_id", unknownContainerId);
```
Willymontaz: The purpose of this container_id field in the map is not obvious. Then reading the rest of the class we king of understand it, but still we need to remember that eventMap is the same fields as the header defined in setUp.
Usually to avoid this and have more readability in tests, I would propose to define header/fsEvents/appEvents factory methods were you can define the fields where a specific value is important and otherwise randomly generated data

readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -105,6 +109,11 @@
     void writeToES(GarmadonMessage msg) {
         String msgType = GarmadonSerialization.getTypeName(msg.getType());
         long timestampMillis = msg.getTimestamp();
+
+        if (GarmadonSerialization.TypeMarker.APPLICATION_EVENT == msg.getType()) elasticSearchCacheManager.addAppEventInCache(msg);
```
Willymontaz: I propose to use a single method -> esCacheManager.cacheEnrichableData(garmadonMsg) and this method has the intelligence to know what event type has what data that can be used in enrichment.

The EsCacheManager itself would be a proxy to several instances of msg specific caches

PR #136, opened 2019-04-25
Set header id on all events

ashangit:
No id is set on NM and RM events which leads to all those events going to same
partition which can lead to slow/unballanced partition
No id is set on APPLICATION_EVENT froom resourcemanager which lead to not have
all events from an app_id on the same partition. We use this assumption on
heuristics readers and on ES readers to enrich data


agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/resourcemanager/RMContextImplEventRunnable.java
```diff
@@ -38,6 +38,7 @@ public void run() {
                 rmContext.getRMApps().forEach((applicationId, rmApp) -> {
                     if (cacheFinishedApp.getIfPresent(applicationId.toString()) == null) {
                         Header header = Header.newBuilder()
+                                .withId(applicationId.toString())
```
gray-matter: Id = ApplicationID, is this expected?

ashangit: Yes it is if field is only used as key for kafka partitioning
in case of app it is app id, in case of other is is currently HeaderUtils.getId

PR #146, opened 2019-05-27
Convert Double.NEGATIVE_INFINITY value on map to Double max negative value

ashangit:
Double.NEGATIVE_INFINITY leads to send events to ES with records_lag_max field equal to -Infinity
which leads to parsing exception on events with this value
Here we will update this value by max negative double value which will be indexable in ES


readers/common/src/main/java/com/criteo/hadoop/garmadon/protobuf/ProtoConcatenator.java
```diff
@@ -91,7 +91,13 @@ protected ProtoConcatenator() {
                 return concatMap;
             },
             (entry, eventMap) -> {
-                eventMap.put(entry.getKey().getName(), getRealFieldValue(entry.getValue()));
+                Object value;
+                if ("DOUBLE".equals(entry.getKey().getType().name())) {
+                    value = Double.NEGATIVE_INFINITY == (Double) entry.getValue() ? -Double.MAX_VALUE : getRealFieldValue(entry.getValue());
```
mchataigner: Use `Double.MIN_VALUE` instead

readers/common/src/test/java/com/criteo/hadoop/garmadon/reader/GarmadonMessageTest.java
```diff
@@ -70,4 +71,18 @@ public void testToMapNormalizedWithPortInfo() {
...
+        long timestampMillis = System.currentTimeMillis();
+        EventHeaderProtos.Header header = EventHeaderProtos.Header.newBuilder().build();
+        FlinkEventProtos.OperatorEvent.Builder operatorEvent = FlinkEventProtos.OperatorEvent.newBuilder()
+            .setRecordsLagMax(Double.NEGATIVE_INFINITY);
+
+        GarmadonMessage message = new GarmadonMessage(GarmadonSerialization.TypeMarker.FLINK_OPERATOR_EVENT, timestampMillis, header, operatorEvent.build(), null);
+
+        Map<String, Object> map = message.toMap(true, true);
+
+        TestCase.assertEquals(-Double.MAX_VALUE, map.get("records_lag_max"));
```
mchataigner: Use `Double.MIN_VALUE` instead

readers/common/src/main/java/com/criteo/hadoop/garmadon/protobuf/ProtoConcatenator.java
ashangit: For NEGATIVE_INFINITY is more closer to -Double.MAX_VALUE than Double.MIN_VALUE

readers/common/src/test/java/com/criteo/hadoop/garmadon/reader/GarmadonMessageTest.java
ashangit: idem

readers/common/src/main/java/com/criteo/hadoop/garmadon/protobuf/ProtoConcatenator.java
mchataigner: -Double.MAX_VALUE is the correct value (as Double.MIN_VALUE is the smallest positive nonzero value of type double.

readers/common/src/test/java/com/criteo/hadoop/garmadon/reader/GarmadonMessageTest.java
mchataigner: -Double.MAX_VALUE is the correct value (as Double.MIN_VALUE is the smallest positive nonzero value of type double.

PR #149, opened 2019-07-03
Add support for key value yarn.tags

mchataigner:
It may be interesting to tag via key value applications running on a hadoop cluster.
Unfortunately, yarn.tags (supported by most frameworks) are a list of string.
This change adds the ability to have special yarn.tags containing a ':' to split key and value.

All those tags will appear in the ApplicationEvent under field user_tags which is a map of
string to string.

It will be exported as nested object on ElasticSearch or Parquet.

### DIFF
--- a/agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/resourcemanager/RMContextImplEventRunnable.java
+++ b/agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/resourcemanager/RMContextImplEventRunnable.java
@@ -66,11 +66,11 @@ public class RMContextImplEventRunnable implements Runnable {
                 .setQueue(rmApp.getQueue());
 
             rmApp.getApplicationTags().stream()
-                    .filter(tag -> YARN_TAGS_TO_EXTRACT.stream().noneMatch(tag::startsWith))
+                    .filter(tag -> YARN_TAGS_TO_EXTRACT.stream().noneMatch(tag::startsWith) && !tag.contains(":"))
                     .forEach(eventBuilder::addYarnTags);
 
             rmApp.getApplicationTags().stream()
-                    .filter(tag -> YARN_TAGS_TO_EXTRACT.stream().anyMatch(tag::startsWith))
+                    .filter(tag -> tag.contains(":") && YARN_TAGS_TO_EXTRACT.stream().anyMatch(tag::startsWith))
                     .map(tag -> {
                         int idx = tag.indexOf(':');
                         String key = tag.substring(0, idx);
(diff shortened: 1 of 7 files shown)


mchataigner:
> please also update readers/elasticsearch/src/main/resources/template.json to add user_tags field description:
> ` "user_tags": { "type": "object" },`
> 
> I'm still not convinced on the real need of this new field. yarn_tags seems enough to me.
> If we move forward with this field is there any way to ensure that number of key won't leads to too many fields being created in ES.

Done

agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/resourcemanager/RMContextImplEventRunnable.java
```diff
@@ -49,23 +61,23 @@ public void sendAppEvent(ApplicationId applicationId, RMApp rmApp) {
...
-                    .filter(tag -> !tag.contains(":"))
+                    .filter(tag -> YARN_TAGS_TO_EXTRACT.stream().noneMatch(tag::startsWith))
                     .forEach(eventBuilder::addYarnTags);
 
             rmApp.getApplicationTags().stream()
-                    .filter(tag -> tag.contains(":"))
+                    .filter(tag -> YARN_TAGS_TO_EXTRACT.stream().anyMatch(tag::startsWith))
                     .map(tag -> {
                         int idx = tag.indexOf(':');
                         String key = tag.substring(0, idx);
```
ashangit: will failed if ":" is not found on tag string as idx will be -1

readers/elasticsearch/src/main/resources/template.json
```diff
@@ -65,6 +65,18 @@
         }
       }
     },
+    "project_name": {
```
ashangit: please apply same config than the one for application_name field. The idea here is to have a project_name.keyword and workflow_name.keyword fields which are fields without split and normalisation 
"application_name": {
      "type": "text",
      "norms": false,
      "index_options": "positions",
      "fields": {
        "keyword": {
          "type": "keyword",
          "ignore_above": 1024
        }
      }
    },

agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/resourcemanager/RMContextImplEventRunnable.java
mchataigner: Done

readers/elasticsearch/src/main/resources/template.json
mchataigner: Done

PR #151, opened 2019-07-04
Some of the JVMStats section contains empty property string leading to bad fields beeing added to ES

ashangit:
If property name is empty we are adding fields section_ to the document.
This leads to some fields like code_, class_... to be added and indexed even if they have empty value.
As they are not usefull we shoulkd just skip them if it is empty

### DIFF
--- a/readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/EventHelper.java
+++ b/readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/EventHelper.java
@@ -31,10 +31,12 @@ class EventHelper {
             } else {
                 Map<String, Object> eventMap = eventMaps.computeIfAbsent(type, s -> EventHelper.initEvent(type));
                 for (JVMStatisticsEventsProtos.JVMStatisticsData.Property property : section.getPropertyList()) {
-                    try {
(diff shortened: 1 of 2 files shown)


readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/EventHelper.java
```diff
@@ -31,10 +31,12 @@ static void processJVMStatisticsData(String type, JVMStatisticsEventsProtos.JVMS
             } else {
                 Map<String, Object> eventMap = eventMaps.computeIfAbsent(type, s -> EventHelper.initEvent(type));
                 for (JVMStatisticsEventsProtos.JVMStatisticsData.Property property : section.getPropertyList()) {
-                    try {
-                        eventMap.put(section.getName() + "_" + property.getName(), Double.parseDouble(property.getValue()));
-                    } catch (NumberFormatException nfe) {
-                        eventMap.put(section.getName() + "_" + property.getName(), property.getValue());
+                    if (!"".equals(property.getName())) {
```
mchataigner: Nit: isEmpty

mchataigner: And maybe add a comment

ashangit:
For me this value can't be null as it is get from proto classes

PR #153, opened 2019-07-19
 Add sum of vcore/memory used and split graph vcore/memory per component on yarn application dashboard

ashangit:


ashangit:
Thanks @jpbempel I badly understood this time metrics.
Will revert this part of the MR

PR #160, opened 2019-07-31
 Add start and finish time to application event and correct vcore/memory computation

ashangit:


agent/src/main/java/com/criteo/hadoop/garmadon/agent/tracers/hadoop/resourcemanager/RMContextImplEventRunnable.java
```diff
@@ -83,23 +82,24 @@ public void sendAppEvent(ApplicationId applicationId, RMApp rmApp) {
                 })
                 .forEach(splitTag -> BUILDERS.get(splitTag[0]).accept(splitTag[1], eventBuilder));
 
+            eventBuilder.setFinalStatus(rmApp.getFinalApplicationStatus().name());
```
Willymontaz: rmApp.getFinalApplicationStatus().name() throws NullPointer if the application is not in FINISHED state

ashangit: This one is never null:
`@Override
  public FinalApplicationStatus getFinalApplicationStatus() {
    // finish state is obtained based on the state machine's current state
    // as a fall-back in case the application has not been unregistered
    // ( or if the app never unregistered itself )
    // when the report is requested
    if (currentAttempt != null
        && currentAttempt.getFinalApplicationStatus() != null) {
      return currentAttempt.getFinalApplicationStatus();
    }
    return createFinalApplicationStatus(this.stateMachine.getCurrentState());
  }`

Willymontaz: ok, confusion from me with the rmAppAttempt.getFinalApplicationStatus() for which we had to patch 
https://github.com/criteo/garmadon/pull/164/files

PR #161, opened 2019-08-14
Don't restart consumption from the beginning when adding a new event type

Willymontaz:


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/kafka/OffsetResetter.java
```diff
@@ -66,32 +66,39 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
+
+        boolean possibleNewEvent = offsetsPerPartition.values().stream().allMatch(offsets -> offsets.contains(UNKNOWN_OFFSET));
+
+        Map<Integer, Long> minOffsetPerPartition = offsetsPerPartition
+            .entrySet()
+            .stream()
+            .collect(Collectors.toMap(Map.Entry::getKey, offsets -> offsets
+                .getValue()
+                .stream()
+                .filter(offset -> offset != UNKNOWN_OFFSET || !possibleNewEvent)
```
gray-matter: What about filtering unknown offsets, then only resetting if the resulting list is empty? This would seem more legible to me.

Willymontaz: agreed that         boolean possibleNewEvent = offsetsPerPartition.values().stream().allMatch(offsets -> offsets.contains(UNKNOWN_OFFSET));
 is not relevant since we only see partitions for this reader.

PR #162, opened 2019-08-19
Remove Safepoints & Locks heuristics from the DrElephant exporter

gray-matter:
These metrics are always red ATM.
They are too complex for users to understand and for us to finetune, so we decide to ignore them for now.

JIRA: LAKECOMP-1303


jpbempel:
Too bad :(
I can adjust the ratios if needed

Willymontaz:
But what is a good ratio for a spark job ? a map reduce job ? a flink job ? 

Also do we have explicit solutions to propose an improvement ?

jpbempel:
It shouldn't be dependent of the framework. They are indicators of code that is not performant (too much locks/contention) and/or generating STW events

I will analyze those figures in the next months to adjust the thresholds / ratios

Willymontaz:
Thank you for looking.

If the framework is the cause of the contention, then it is framework dependant right ? And users won't be able to do anything then.

PR #163, opened 2019-08-19
Report latest HDFS reader committed event timestamp per event type in Prometheus

Willymontaz:


readers/hdfs/src/test/java/com/criteo/hadoop/garmadon/hdfs/writer/PartitionedWriterTest.java
```diff
@@ -319,7 +320,12 @@ public void heartbeatWithNoMessage() throws IOException {
 
         writer.heartbeat(partition, offset);
 
-        verify(consumer, times(1)).write(argument.capture(), any(Offset.class));
+        verify(consumer, times(1)).write(timestampCaptor.capture(), argument.capture(), any(Offset.class));
+
+        Long timestamp = timestampCaptor.getValue();
+        //timestamp should be now, testing a 10 second range is enough for the purpose
+        long now = System.currentTimeMillis();
+        assertTrue(timestamp - 10000 < now && now < timestamp + 10000);
```
gray-matter: assertEquals(now, timestamp, 10000)

PR #169, opened 2019-09-06
[HDFS Reader] Use timestamp of the current day as the default value for latest_committed_timestamp

Willymontaz:
Details: we faced the case of constant failures during file committing, that induced restart of reader instances
Those restart led to think that latest committed timestamp was making progress, which was fake.

Using a constant value for the day, across restart will help to catch such issue.

### DIFF
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
@@ -18,6 +18,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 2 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
```diff
@@ -196,10 +197,9 @@ private double getLatestCommittedTimestamp() {
         //there are cases for which we won't find a value for the latest committed timestamp
         // - the first time this code goes in, no file has the correct metadata
         // - for a new event type, we have no history too, so no value
-        //By using the default value 'now' rather than 0, we prevent firing unnecessary alerts
-        //However, if there is an actual problem and the reader never commits, it will eventually fire
-        //an alert.
-        long defaultValue = System.currentTimeMillis();
+        // - if started and no file has been create for the day
```
gray-matter: "created"

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/ProtoParquetWriterWithOffset.java
```diff
@@ -196,10 +197,9 @@ private double getLatestCommittedTimestamp() {
...
+        //Using the timestamp of today midnight prevents unnecessary alerts
+        long defaultValue = LocalDateTime.now().toLocalDate().atStartOfDay().toEpochSecond(ZoneOffset.UTC) * 1000;
```
mk-dhia: Why we need a changing name. maybe just make sure that alerting do not use the file name as source of information. and use a constant name that at least will be clear that file is created because there is nothing to commit

gray-matter: Not sure where you see a file name. This is the timestamp for the monitoring/alerting part, it's not used for files naming (cf tests).

Willymontaz:
The new default value is not great either. Maybe we should just put something like 0 so we can clearly identified this corner case. I prefer having a false alert once in a while (should be only for new events and first time the commit was deployed) that disappears at the first file correctly committed or stays if there is really an issue.

PR #176, opened 2019-09-17
Reconnect to the SQL DB if reader failed to execute statement

ashangit:


readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/helper/ReaderUtils.java
```diff
@@ -0,0 +1,39 @@
...
+    protected ReaderUtils() {
+        throw new UnsupportedOperationException();
+    }
+
+    public static boolean retryAction(Runnable action, String exceptionStr, Runnable actionFailure) {
+        final int maxAttempts = 5;
+
+        for (int retry = 1; retry <= maxAttempts; ++retry) {
+            try {
+                action.run();
```
Willymontaz: use Callable for action. That allows to return what is potentially generated by the action (cna serve more use cases than Runnable and return boolean) + Callable can throw Exception. So you don't have to catch IOException and wrap in RuntimeException for instance

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/helper/ReaderUtils.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final Runnable EMPTY_ACTION = () -> {
+    };
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ReaderUtils.class);
+
+    protected ReaderUtils() {
+        throw new UnsupportedOperationException();
+    }
+
+    public static boolean retryAction(Runnable action, String exceptionStr, Runnable actionFailure) {
```
Willymontaz: [OPT] actionFailure could be Consumer<Throwable> to allow callee to do something with the exception

readers/heuristics/src/test/java/com/criteo/hadoop/garmadon/heuristics/HeuristicsResultDBTest.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    private void waitUntilDbStopped() {
+        while (true) {
+            try {
+                new ServerSocket(port).close();
+                break;
+            } catch (IOException e) {
+                try {
+                    Thread.sleep(1L);
```
Willymontaz: sleep(1000) ?

readers/heuristics/src/test/java/com/criteo/hadoop/garmadon/heuristics/HeuristicsResultDBTest.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Test
+    public void reconnectOnFailure() {
+        String appId = "appId";
+        String detail = "detail";
+        HeuristicsResultDB db = spy(new HeuristicsResultDB(dbConfiguration));
+        HeuristicResult heuristicResult = new HeuristicResult(appId, "attemptId", GCCause.class, 1, 1);
+        heuristicResult.addDetail(detail, "value", "This is a new detail");
+        new Thread(() -> {
```
Willymontaz: why stopping in another thread if we waitUntilDbStopped ? There is something I don't get

readers/heuristics/src/test/java/com/criteo/hadoop/garmadon/heuristics/HeuristicsResultDBTest.java
ashangit: Wanted to be sure that I start to try inserting data immediately after the db is topped to avoid having the db restarted before inserting. Otherwise won't see any call to retry.
But I can update it to a bigger value (10 or a 100)

readers/heuristics/src/test/java/com/criteo/hadoop/garmadon/heuristics/HeuristicsResultDBTest.java
ashangit: True moving the stop out of th thread and keep the start in it

PR #177, opened 2019-09-18
Try to solve Hdfs reader issue

mchataigner:
Catch exception during onPartitionAssigned and onPartitionRevoked and log errors
Activate some Trace + Info hdfs logging for DFSClient and socket stack.


readers/hdfs/src/main/resources/logback.xml
```diff
@@ -9,6 +9,11 @@
     <logger name="org.apache.kafka.common.utils.AppInfoParser" level="INFO" />
     <logger name="org.apache.kafka.clients.consumer.ConsumerConfig" level="INFO" />
 
+    <logger name="org.apache.hadoop.hdfs.RemoteBlockReader2" level="TRACE" />
```
Willymontaz: To change log level, change configuration on mesos-configuration

Willymontaz: This one is only used for local debugging and won't be taken into account

mchataigner: I put this in mesos-configuration.
I don't understand your second statement.

PR #178, opened 2019-09-25
Increase sleep duration between DB connection retry

ashangit:


readers/heuristics/src/main/java/com/criteo/hadoop/garmadon/heuristics/HeuristicsResultDB.java
```diff
@@ -136,6 +136,6 @@ public void executeUpdate(Callable action, String exceptionStr) {
                     initConnectionAndStatement(dbConfiguration);
                 } catch (Exception ignored) {
                 }
-            });
+            }, 15000L);
```
Willymontaz: would make sense to define a property with default value

PR #180, opened 2019-10-09
[HDFS-READER] Solve deadlock issue

mchataigner:
run heartbeat and expirer in the same thread as garmadon reader to avoid deadlock issues.
reduce the scope of some synchronized blocks.

### DIFF
--- a/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
+++ b/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
@@ -18,6 +18,8 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
+import java.time.LocalDateTime;
+import java.time.temporal.TemporalAmount;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -53,9 +55,9 @@ public final class GarmadonReader {
     private boolean reading = false;
 
     private GarmadonReader(Consumer<String, byte[]> kafkaConsumer, List<GarmadonMessageHandler> beforeInterceptHandlers,
-                           Map<GarmadonMessageFilter, GarmadonMessageHandler> listeners, int id) {
+                           Map<GarmadonMessageFilter, GarmadonMessageHandler> listeners, List<RecurrentAction> recurrentActions, int id) {
         this.cf = new CompletableFuture<>();
-        this.reader = new Reader(kafkaConsumer, beforeInterceptHandlers, listeners, cf);
+        this.reader = new Reader(kafkaConsumer, beforeInterceptHandlers, listeners, recurrentActions, cf);
         this.id = id;
     }
 
@@ -86,9 +88,31 @@ public final class GarmadonReader {
         }
     }
 
+    public static class RecurrentAction {
+        private final TemporalAmount period;
+        private final Runnable action;
+
+        private LocalDateTime nextRunTimestamp;
+
+        public RecurrentAction(Runnable action, TemporalAmount period) {
+            this.action = action;
+            this.period = period;
+            nextRunTimestamp = LocalDateTime.now();
+        }
+
+        public void run() {
+            LocalDateTime current = LocalDateTime.now();
+            if (current.isAfter(nextRunTimestamp)) {
+                nextRunTimestamp = current.plus(period);
+                action.run();
+            }
+        }
+    }
+
     protected static class Reader implements Runnable {
 
         private final SynchronizedConsumer<String, byte[]> consumer;
+        private final List<RecurrentAction> recurrentActions;
         private final CompletableFuture<Void> cf;
         private final List<GarmadonMessageHandler> beforeInterceptHandlers;
         private final Map<GarmadonMessageFilter, GarmadonMessageHandler> listeners;
@@ -99,11 +123,12 @@ public final class GarmadonReader {
         private volatile boolean keepOnReading = true;
 
         Reader(Consumer<String, byte[]> consumer, List<GarmadonMessageHandler> beforeInterceptHandlers, Map<GarmadonMessageFilter,
-            GarmadonMessageHandler> listeners, CompletableFuture<Void> cf) {
+            GarmadonMessageHandler> listeners, List<RecurrentAction> recurrentActions, CompletableFuture<Void> cf) {
             this.consumer = SynchronizedConsumer.synchronize(consumer);
             this.beforeInterceptHandlers = beforeInterceptHandlers;
             this.listeners = listeners;
             this.filters = listeners.keySet();
+            this.recurrentActions = recurrentActions;
             this.cf = cf;
         }
 
@@ -114,6 +139,7 @@ public final class GarmadonReader {
                 LOGGER.info("initialize reading");
 
                 while (keepOnReading) {
+                    recurrentActions.forEach(RecurrentAction::run);
                     readConsumerRecords();
                 }
             } catch (Exception e) {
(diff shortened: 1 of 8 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/kafka/PartitionsPauseStateHandler.java
```diff
@@ -44,18 +44,21 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
     }
 
     public void pause(Class pausingEvent) {
-        synchronized (consumer) {
-            if (pausedEvents.isEmpty() && !currentlyAssignedPartitions.isEmpty()) consumer.pause(currentlyAssignedPartitions);
-
+        if (pausedEvents.isEmpty() && !currentlyAssignedPartitions.isEmpty()) {
```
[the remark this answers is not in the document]
mchataigner: hmm, you're right.
However, onPartitionAssigned andonPartitionRevoked are called inside consumer.poll which is currently doing a synchronized(consumer).
So it should be safe, but I will modify the code.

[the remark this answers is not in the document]
Willymontaz: I just jumped in. onPartitionsRevoked and onPartitionsAssigned are called by the same thread than does the consumer.poll (actually onPartitionsRevoked and onPartitionsAssigned are part of the consumer.poll process). 
Was that the concern ?

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HeartbeatConsumer.java
```diff
@@ -41,18 +41,7 @@ public HeartbeatConsumer(Collection<PartitionedWriter<MESSAGE_KIND>> writers, Te
...
-                        Offset offset = partitionOffset.getValue();
-                        Offset latestHeartbeat = latestHeartbeats.get(partition);
-
-                        if (!offset.equals(latestHeartbeat)) {
-                            latestHeartbeats.put(partition, offset);
-                            writers.forEach(writer -> writer.heartbeat(partition, offset));
-                        }
-                    }
-                }
+                run();
```
Willymontaz: I thought this commit would avoid running HeartbeatConsumer and Expirer in dedicated threads. Did I miss something ?

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HeartbeatConsumer.java
```diff
@@ -68,6 +57,21 @@ public void start(Thread.UncaughtExceptionHandler uncaughtExceptionHandler, Stri
         runningThread.start();
     }
 
+    public void run() {
+        synchronized (latestPartitionsOffset) {
```
Willymontaz: we should not synchronise latestPartitionOffset anymore. It will always be called by the consumer thread (if we remove Heartbeat consumer thread of course)

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/PartitionedWriter.java
```diff
@@ -285,7 +285,7 @@ public Expirer(Collection<PartitionedWriter<MESSAGE_KIND>> writers, TemporalAmou
         public void start(Thread.UncaughtExceptionHandler uncaughtExceptionHandler, String name) {
             runningThread = new Thread(() -> {
                 while (!Thread.currentThread().isInterrupted()) {
-                    writers.forEach(PartitionedWriter::expireConsumers);
+                    run();
```
Willymontaz: Same remark than for the Heartbeat consumer + we should also remove all synchronization on the hashmap of PartitionedWriter

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/kafka/PartitionsPauseStateHandler.java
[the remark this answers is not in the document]
jpbempel: hmm maybe I was not clear enough :)
Methods `onParitionRevoked` &  `onPartitionsAssigned` are not the concerned. If those methods are called by the consumer.poll process or whatever and hence under `synchronized (consumer)` does not make it more safe. It may just say that why are you synchronizing on consumer inside those methods then...

My concern is about `pause`/`resume` methods in which we are reading into `currentlyAssignedPartitions` instance without any read barrier because of @mchataigner change.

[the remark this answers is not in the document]
Willymontaz: Ok. What's misleading is that there should not be any synchronized anywhere anymore since pause/resume are also intended to be called from the same thread.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HeartbeatConsumer.java
mchataigner: done

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/offset/HeartbeatConsumer.java
mchataigner: done

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/PartitionedWriter.java
mchataigner: done

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/PartitionedWriter.java
```diff
@@ -282,45 +280,8 @@ public Expirer(Collection<PartitionedWriter<MESSAGE_KIND>> writers, TemporalAmou
...
-
-                return CompletableFuture.supplyAsync(() -> {
-                    try {
-                        runningThread.join();
-                    } catch (InterruptedException e) {
-                        LOGGER.info("Exception caught while waiting for expirer thread to finish", e);
-                    }
-
-                    return null;
-                }).thenRun(() -> writers.forEach(PartitionedWriter::close));
```
sun77: Don't you forget to close the PartitionedWriter ?

Willymontaz:
Please add "[HDFS-READER]" in the commit message once you merge

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/writer/PartitionedWriter.java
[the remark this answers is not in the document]
mchataigner: done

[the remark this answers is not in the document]
mchataigner: done

readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
```diff
@@ -114,17 +142,33 @@ public void run() {
...
                     readConsumerRecords();
                 }
             } catch (Exception e) {
                 LOGGER.error("unexpected exception while reading", e);
                 cf.completeExceptionally(e);
             }
 
+            final RuntimeException firstExceptionDuringPostReadingActions = new RuntimeException("failed to run successfuly post reading actions");
+
+            postReadingActions.forEach(r -> {
```
Willymontaz: GarmadonReader already uses a CompletableFuture to notify user code that the read is complete. For now, it is only accessible via call to start and stop method. To close all partitioned writers I propose the following:

- Add a method in garmadon reader returning the completable future (field cf)
- In ReaderFactory.create method, before returning the reader, call the method above, and chain the CompletableFuture with whenCompleted. At this place of the code, in the whenCompleted callback, you can close all partitioned writers.

mchataigner: done

PR #184, opened 2019-11-16
ES reader requires a synchronized consumer

mchataigner:
As ES reader processes messages in bulk in a separated thread, committing offsets will occur in this other thread and not reader thread. As it touches kafka consumer, it may cause some race condition.
Putting back Synchronized consumer for ES solves this issue.

### DIFF
--- a/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
+++ b/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
@@ -261,7 +261,7 @@ public final class GarmadonReader {
         }
     }
 
-    public static class SafeGarmadonConsumer<K, V> {
+    public static final class SafeGarmadonConsumer<K, V> {
         private final Consumer<K, V> consumer;
 
         private SafeGarmadonConsumer(Consumer<K, V> consumer) {
--- a/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
+++ b/readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
@@ -18,6 +18,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
+import java.time.Duration;
 import java.time.LocalDateTime;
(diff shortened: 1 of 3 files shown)


readers/common/src/main/java/com/criteo/hadoop/garmadon/reader/GarmadonReader.java
```diff
@@ -258,6 +259,70 @@ int get() {
...
+    private static final class SynchronizedConsumer<K, V> implements GarmadonConsumer<K, V> {
+
+        private final Consumer<K, V> consumer;
+
+        private SynchronizedConsumer(Consumer<K, V> consumer) {
+            this.consumer = consumer;
+        }
+
+        public synchronized ConsumerRecords<K, V> poll(long timeout) {
+            synchronized (consumer) {
```
jpbempel: Why the double synchronization?

mchataigner: it was simple copy paste of how the code was before.
I removed the duplicated synchronized.
Thanks for the review.

Willymontaz: For the record, it was because in HdfsReader we were synchronizing on the consumer instance. 
Allowing to create a consumer outside of GarmadoReader scope was necessary to access other methods of the consumer before starting the read process (subscribe with listeners). 
SynchronizedConsumer is supposed to be used only inside GarmadonReader to provide CommittableOffset facility. Since ReaderBuilder now accepts externally created consumer it is better to synchronize on the instance instead of the class like you did.

Willymontaz: It does not justify the double synchronize but I guess the person who did the commit was not confident enough ad did not want to break an existing behavior.

Willymontaz:
We should always use the synchronized consumer. This way GarmadonReader is clean with its contract. You can read and commit from any thread, that is the facility provided by GarmadonReader. There is no contention at all on poll and commit so there is no perf penalty.
Synchronizing on consumer instance created outside GarmadonReader scope would be the matter of the user code that does something fancy outside GarmadonReader's control.
In the end less code and less reasoning for all cases.

mchataigner:
@Willymontaz done

PR #188, opened 2019-12-11
[HDFS READER] Fix possible connection leak due to readers not closed

Willymontaz:


mchataigner:
I would recommend to push a patch to parquet upstream.

PR #196, opened 2020-06-10
Add usage of ES index lifecycle on top of garmadon indexes

ashangit:
Index management will now rely on ES ILM mechanism (https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) to avoid having to maintain different external processes


ashangit:
@Willymontaz @sun77 plz review

PR #198, opened 2020-10-30
Add support to https scheme on elasticsearch reader

ashangit:


readers/elasticsearch/src/main/java/com/criteo/hadoop/garmadon/elasticsearch/ElasticSearchReader.java
```diff
@@ -252,7 +254,12 @@ private static BulkProcessor setUpBulkProcessor(ElasticsearchConfiguration elast
         putGarmadonTemplate(esClient, elasticsearch);
         putGarmadonLifecyclePolicy(esClient, elasticsearch);
 
-        Sniffer sniffer = Sniffer.builder(esClient.getLowLevelClient()).build();
+        NodesSniffer nodesSniffer = new ElasticsearchNodesSniffer(
+            esClient.getLowLevelClient(),
+            ElasticsearchNodesSniffer.DEFAULT_SNIFF_REQUEST_TIMEOUT,
+            ElasticsearchNodesSniffer.Scheme.valueOf(elasticsearch.getScheme().toUpperCase()));
+
+        Sniffer sniffer = Sniffer.builder(esClient.getLowLevelClient()).setNodesSniffer(nodesSniffer).build();
```
geobeau, line 262: Maybe it would make sense to have a feature flag: activateSniffer, that would let the operator disable the sniffer for a few hours before reenabling it. It may also benefit non-Criteo users that don't want to use the sniffer at all

ashangit: Not linked to that changed, could be done in another PR

Willymontaz (approved):

PR #201, opened 2020-11-13
Send messages to any partitions but some configured types

Willymontaz:
Sending all events for an application id to all kafka partitions creates a sharding issue and can at worst DDOS a kafka partition master

This commit forces all events to be sent to any partitions. Making so, we lose the possibility to enrich some events with information from other events (such as elastic search reader does).
To cope with this issue, this commit also introduce the possibility to send some specific type of events to all partition.

Sending events to all partitions should be reserved for low frequency events. By default, no event are sent to all partitions.

### DIFF
--- a/forwarder/src/main/java/com/criteo/hadoop/garmadon/forwarder/kafka/KafkaService.java
+++ b/forwarder/src/main/java/com/criteo/hadoop/garmadon/forwarder/kafka/KafkaService.java
@@ -26,7 +26,7 @@ public class KafkaService {
                 send(new ProducerRecord<>(topic, partition.partition(), null, value));
             });
         } else {
-            send(new ProducerRecord<>(topic, null, value));
+            send(new ProducerRecord<>(topic, value));
         }
     }
 
--- a/forwarder/src/main/java/com/criteo/hadoop/garmadon/forwarder/kafka/KafkaService.java
+++ b/forwarder/src/main/java/com/criteo/hadoop/garmadon/forwarder/kafka/KafkaService.java
@@ -20,9 +20,17 @@ public class KafkaService {
         this.producer = new KafkaProducer<>(properties);
     }
 
-    public void sendRecordAsync(String key, byte[] value) {
-        ProducerRecord<String, byte[]> record = new ProducerRecord<>(topic, key, value);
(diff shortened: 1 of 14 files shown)


forwarder/src/test/java/com/criteo/hadoop/garmadon/forwarder/handler/EventHandlerTest.java
```diff
@@ -29,16 +27,23 @@
...
+        broadcastedTypes.add(broadCastedType2);
+        EventHandler eventHandler = new EventHandler(broadcastedTypes);
         channel.get().pipeline().addLast(eventHandler);
 
         GarmadonSerialization.register(TestEvent.class, Integer.MAX_VALUE, "TestEvent", event -> event.bytes, TestEvent::new);
     }
 
     @Test
-    public void EventHandler_should_read_event_according_top_protocol() throws TypeMarkerException, SerializationException {
+    public void EventHandler_should_read_non_braodcasted_event() throws TypeMarkerException, SerializationException {
```
gray-matter, line 46: Typo on broadcasted

forwarder/src/main/java/com/criteo/hadoop/garmadon/forwarder/kafka/KafkaService.java
```diff
@@ -20,9 +20,17 @@ public KafkaService(Properties properties) {
...
 
-    public void sendRecordAsync(String key, byte[] value) {
-        ProducerRecord<String, byte[]> record = new ProducerRecord<>(topic, key, value);
+    public void sendRecordAsync(byte[] value, boolean broadcasted) {
+        if (broadcasted) {
+            producer.partitionsFor(topic).forEach(partition -> {
+                send(new ProducerRecord<>(topic, partition.partition(), null, value));
+            });
+        } else {
+            send(new ProducerRecord<>(topic, null, value));
```
gray-matter, line 29: OPT: Could be replaced by new ProducerRecord<>(topic, value)

gray-matter (changes requested):
Minor changes, can approve instead if needed

gray-matter (approved):

PR #203, opened 2020-11-23
[HDFS READER] Reduce the amount of Hive queries

gray-matter:
* There was a bug executing succesful queries 5 times
* Execution of create table and create partition queries is now cached so that no two identical two of them will be performed.
Limitation is this is in memory, but this should be sufficient.
* Splitted HiveClient between execution and abstraction levels; execute as well as some methods that should be private are still
in the public interface of HiveQueryExecutor because fixing this would cause scope creep

### DIFF
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/hive/HiveClient.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/hive/HiveClient.java
@@ -23,7 +23,7 @@ public class HiveClient {
     private final Set<String> createdTables = new HashSet<>();
 
     public HiveClient(HiveQueryExecutor executor, String location) throws SQLException {
-        this.createdPartitions = Collections.newSetFromMap(new LinkedHashMap<Pair<String, String>, Boolean>(){
+        this.createdPartitions = Collections.newSetFromMap(new LinkedHashMap<Pair<String, String>, Boolean>() {
             protected boolean removeEldestEntry(Entry<Pair<String, String>, Boolean> eldest) {
                 return size() > 100;
             }
--- a/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/hive/HiveClient.java
+++ b/readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/hive/HiveClient.java
@@ -1,7 +1,10 @@
 package com.criteo.hadoop.garmadon.hdfs.hive;
 
 import com.google.common.annotations.VisibleForTesting;
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.Map.Entry;
 import java.util.Set;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.parquet.schema.MessageType;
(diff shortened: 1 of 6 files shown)


readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/hive/HiveClient.java
```diff
@@ -1,79 +1,51 @@
...
 
 public class HiveClient {
     private static final Logger LOGGER = LoggerFactory.getLogger(HiveClient.class);
 
-    private final String database;
-    private final String jdbcUrl;
-    private Connection connection;
-    private Statement stmt;
+    private final HiveQueryExecutor executor;
+    private final Set<Pair<String, String>> createdPartitions = new HashSet<>();
```
Willymontaz, line 19: even if it takes a lot of time to fill memory with these, we should find a way to clean old entries

Willymontaz (changes requested):

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/hive/HiveClient.java
gray-matter: I was unsure whether this was premature optimization, but since it seems fairly simple to implement, I did.

readers/hdfs/src/main/java/com/criteo/hadoop/garmadon/hdfs/hive/HiveClient.java
```diff
@@ -16,10 +19,15 @@
...
 
     private final HiveQueryExecutor executor;
-    private final Set<Pair<String, String>> createdPartitions = new HashSet<>();
+    private final Set<Pair<String, String>> createdPartitions;
     private final Set<String> createdTables = new HashSet<>();
 
     public HiveClient(HiveQueryExecutor executor, String location) throws SQLException {
+        this.createdPartitions = Collections.newSetFromMap(new LinkedHashMap<Pair<String, String>, Boolean>(){
+            protected boolean removeEldestEntry(Entry<Pair<String, String>, Boolean> eldest) {
+                return size() > 100;
```
Willymontaz, line 28: Cool, I did not know about this method !

I feel like 100 is enough but could you just explain your rational for this number ? Because we could easily go up to thousands to avoid any surprise as the goal is mostly to prevent out-of-memory issues after days of run

Willymontaz: However I agree that is it best effort to prevent to many requests...

gray-matter: AFAIR, there are 8 event types and usually a 2-day window. So theoretically even 16 should be enough. It's a protective cache anyway, so it doesn't hurt if a query slips through. We mostly want to avoid the bulk, so even 8 could be sufficient.

gray-matter: So there's no strong rationale for 100. I could have picked a power of 2 as well, wouldn't have more sense :smile:

Willymontaz: lol

Willymontaz (approved):

PR #205, opened 2021-02-23
Allow garmadon ElasticSearch reader to read from garmadon topic kafka

younes-b:


gray-matter (approved):
OPT: typo in ElasticSearch (you may change this when merging)

PR #212, opened 2021-12-17
Force inclusion of hive jdbc dependencies used outside of hive jdbc in the uber jar

Willymontaz:


Willymontaz:
I missed some stuff, needs an update

PR #213, opened 2022-02-04
Add possibility to run CI manually in GitHub Actions

jlundqvist-criteo:


.github/workflows/build.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  workflow_dispatch:
```
gray-matter, line 11: This is quite mysterious, would you have some link to a doc explaining why this does what you want to?

jlundqvist-criteo: Yes indeed. This PR was mostly to try out the CI outside the other open PR (to make sure the issue lies with the changes in that PR and not with the CI in general), but this will be a nice feature to have in any case.

Basically setting a trigger on workflow_dispatch allows us to manually launch the workflow from the github UI: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

PR #230, opened 2024-09-04
Send a subset of Spark properties on application start and end

ghost:
Log basic driver and executor properties.
Sending an event when the application starts and when it ends will allow to catch configs that may have changed during execution, and differ from those sent in the spark-submit.

### DIFF
--- a/frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
+++ b/frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
@@ -4,6 +4,7 @@ import com.criteo.hadoop.garmadon.TriConsumer;
 import com.criteo.hadoop.garmadon.event.proto.SparkEventProtos;
 import com.criteo.hadoop.garmadon.schema.enums.State;
 import com.criteo.hadoop.garmadon.schema.events.Header;
+import org.apache.spark.SparkConf;
 import org.apache.spark.scheduler.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
```diff
@@ -77,6 +78,43 @@ private void sendExecutorStateEvent(long time, State state, String executorId, S
...
+    private void sendApplicationStateEvent(long time, State state) {
+        SparkConf conf = new SparkConf();
+        String deployMode = conf.get("spark.submit.deployMode", "unknown");
+
+        int driverMemoryMb = (int) conf.getSizeAsMb("spark.driver.memory", "0");
+        int driverMemoryOverheadMb = (int) conf.getSizeAsMb("spark.driver.memoryOverhead", "0");
+        int driverCores = (int) conf.getInt("spark.driver.cores", 0);
+
+        int executorMemoryMb = (int) conf.getSizeAsMb("spark.executor.memory", "0");
+        int executorMemoryOverheadMb = (int) conf.getSizeAsMb("spark.executor.memoryOverhead", "0");
```
Willymontaz, line 90: we should also get the value of spark.executor.memoryOverheadFactor that will be used if memoryOverhead is not set

frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
```diff
@@ -77,6 +78,43 @@ private void sendExecutorStateEvent(long time, State state, String executorId, S
         this.eventHandler.accept(time, buildOverrideHeader(executorId), executorStateEvent.build());
     }
 
+    private void sendApplicationStateEvent(long time, State state) {
```
Willymontaz, line 81: I would give a try using the constants in core/src/main/scala/org/apache/spark/internal/config/package.scala to access the conf values because they can provide the default values as well

Willymontaz (changes requested):

frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
aroville-criteo: Those constants are private, I cannot reference them from the listener

frameworks/spark/src/main/java/com/criteo/hadoop/garmadon/spark/listener/GarmadonSparkListener.java
aroville-criteo: Done

PR #237, opened 2025-03-24
Expose Forwarder client host information

ghost:
In the forwarder, log client host address and port, along with the protool version used in the greeting.

This allows to keep track of the client version when the protocol changes, as we cannot get metrics directly from clients.


Willymontaz (changes requested):
I wonder one possible problem here -> there will be a single metric sent to prometheus by client, and at the same time a very high cardinality of client. High cardinality ued to be bad for prometheus, not sure this stays true but what happens for an old point sent once after prometheus restarted ? (or after the forwarder instance is restarted)

aroville-criteo:
I updated the change to only log new connections instead of having a matrics that could create cardinality issues in prometheus

Willymontaz (approved):

